PR #2, opened 2015-05-28
types: use int64 when we expect large integers

vincentbernat:
On 32-bit arch, int is usually like int32 and is too small for some
types. For example, the size in bytes doesn't fit in 32 bits. Change
some types to int64 when it makes sense.


pyr:
Thanks @vincentbernat ! :rocket: :guitar:

PR #4, opened 2016-05-02
Add Affinity Group support

runseb:
* This creates several types to define Affinity Groups (AF).
* It modifies the Topology and Machine Profile types to add AF.
* It adds an `affinitygroup.go` file for create and delete of AF.
* This modifies `vm.go` to set the AF on VM creation.

Note that if no AF are defined in the profile the VM creation will work.
You can pass multiple AF in the profile.

The example `exo.go` has also been modified to start the VM in a default _egoscale_ AF. Which is created if it does not exist.

### DIFF
--- a/exo.go
+++ b/exo.go
@@ -52,11 +52,34 @@ func main() {
 		sgid = resp.Id
 	}
 
+
+	agid, present := topo.AffinityGroups["egoscale"]
+	if !present {
+		//Affinity Group Create is an async call
+		jobid, err := client.CreateAffinityGroup("egoscale")
+
+		var resp *egoscale.QueryAsyncJobResultResponse
+		for i := 0; i <= 10; i++ {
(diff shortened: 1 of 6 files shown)


exo.go
```diff
@@ -52,11 +52,34 @@ func main() {
...
+		}
+	}
+	fmt.Printf("Affinity Group ID :%v\n",agid)
+
 	profile := egoscale.MachineProfile{
 		Template: topo.Images["ubuntu-14.04"][10],
 		ServiceOffering: topo.Profiles["large"],
 		SecurityGroups: []string{ sgid,},
 		Keypair: topo.Keypairs[0],
+		AffinityGroups: []string{"egoscale"},
```
pyr: unsure there should be a default one

pyr:
LGTM. Thanks!

PR #5, opened 2016-05-23
Added features for terraform plugin

runseb:
A few added features/tweaks to be used in the terraform plugin


pyr:
Thanks! This is quite helpful.

PR #6, opened 2016-09-17
Add support for full retrieval of SecurityGroup data

cab105:
In order to allow for idempotent support of passing ingress/egress rules with security groups for the Exoscale terraform plugin, I needed to modify the egoscale library to return more data than just the Security Group Id based off the name.


pyr:
Thanks @cab105 !

cab105:
yw

src/egoscale/topology.go
```diff
@@ -142,7 +142,7 @@ func (exo *Client) GetImages() (map[string]map[int]string, error) {
 	images = make(map[string]map[int]string)
 
 	params := url.Values{}
-	params.Set("templatefilter", "featured")
+	params.Set("templatefilter", "executable")
```
marcaurele: Not a good idea

PR #9, opened 2017-01-13
Clean version of nested SG support

retrack:


greut:
This PR seems to be already part of the current code.

PR #11, opened 2017-08-17
Fix dependency

tamalsaha:
I am trying to vendor this repo as part of my https://github.com/xenolf/lego dependency. But the vendoring process fails because of this bad import.

I am using the new golang/dep tool to vendor.

### DIFF
--- a/exo.go
+++ b/exo.go
@@ -1,10 +1,11 @@
 package main
 
 import (
-	"egoscale"
-	"fmt"
 	"flag"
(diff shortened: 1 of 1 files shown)


marcaurele:
@tamalsaha  It should be fixed now, using the last version https://github.com/exoscale/egoscale/commit/325740036187ddae3a5b74be00fbbc70011c4d96 which includes other changes.

Closing the PR

PR #12, opened 2017-09-12
Renaming

marcaurele:
- Remove src/egoscale path
- go fmt
- Remove Makefile and follow `go get` standard


marcaurele:
Superseding #11

PR #16, opened 2017-12-07
fix: docker-machine driver expects lowercase strings

greut:
https://github.com/exoscale/docker-machine/blob/master/drivers/exoscale/exoscale.go#L311-L321

also line 181, meh...


greut:
Merged it for docker-machine (and the next big thing®)

PR #24, opened 2017-12-21
groups: creation using a profile and some reflection

greut:
Addition of `CreateSecurityGroup` using a `SecurityGroupProfile`. We are bound to document the whole CloudStack API in Go structs... but it already a little bit nicer.

And super usage of the `reflect` package to create POST requests based on a `struct`


greut:
List of the supported calls so far... https://github.com/exoscale/egoscale/wiki/api

request.go
```diff
@@ -190,3 +206,85 @@ func (exo *Client) DetailedRequest(uri string, params string, method string, hea
...
+
+// prepareValues uses a profile to build a POST request
+func prepareValues(profile interface{}) (*url.Values, error) {
+	params := url.Values{}
+	value := reflect.ValueOf(profile)
+	typeof := reflect.TypeOf(profile)
+	for typeof.Kind() == reflect.Ptr {
+		typeof = typeof.Elem()
+		value = value.Elem()
+	}
```
marcaurele: For my know-how, what is the intent of this loop?

greut: `interface{}` could be the `struct` I want to work on, but also a `*struct`, or a `**struct`, and so on.

greut:
too big to parse

PR #25, opened 2017-12-28
refactoring: big and long

greut:
- using reflect to build requests
- restore the old API (marking it as [Deprecated](https://rakyll.org/deprecated/))
- add documentation, preview: http://159.100.241.210/

The way you should use `egoscale` is completely different...

```golang
client = NewClient(endpoint, token, secret)

resp, err := client.Request(&CreateSecurityGroup{
	Name: "Load balancer",
	Description: "Opens HTTP/HTTPS ports from the outside world",
})
securityGroup := resp.(*CreateSecurityGroupResponse).SecurityGroup

err := client.BooleanRequest(&DeleteSecurityGroup{
	ID: securityGroup.ID,
})
```

You can do everything using the four following methods:

- `client.Request(*req, *resp)`
- `client.AsyncRequest(*req, *resp, AsyncInfo)`
- `client.BooleanRequest(*req)`
- `client.BooleanAsyncRequest(*req, AsyncInfo)`

Implementing now features is a matter of translating the `listApis` response to Go `struct`s

:warning: Not affected and not tested, `dns.go` :warning:


vincentbernat:
Is the old API still working? As there are changes to structures, it seems likely the old API doesn't work. So, maybe instead of maintaining a deprecated API, we should just drop it.

vincentbernat:
What about:

```go
client = NewClient(endpoint, token, secret)
if resp, err := client.Request(&CreateSecurityGroupRequest{
	Name: "Load balancer",
	Description: "Opens HTTP/HTTPS ports from the outside world",
}); err != nil {
  resp := resp.(*CreateSecurityGroupResponse)
  /*...*/
}
```

*Request could get a function to return an empty *Response.

In Go, there are cases where you provide the recipient for the result as argument (like Unmarshal), but in this case, the default values are used and there is no way for Unmarshal to know the recipient structure beforehand.

greut:
@vincentbernat Dropping the old calls already? Why not, they will be.

To my understanding, it's hard to have types on the fly.

```golang
type Request interface {
    Command() string
    Response() interface{}
}

type FooRequest struct {}
type FooResponse struct{}

func (req *FooRequest) Command() string {
    return "foo"
}
func (req *FooRequest) Response() interface{} {
    return new(FooResponse)
}
```

It doesn't make the documentation clearer... Making them private is a solution, documentation-wise.

greut:
the kiss-cool effect of that is that you may be able to `resp.(*ErrorResponse)`

vincentbernat:
Currently, you also have an interface (as second argument), so it seems equivalent. It just seems more natural to use a return value.

As for making `Command` and `Response` private, yes, this should be.

greut:
Another change, all `...Request` implements `Command` or `AsyncCommand`. That way, at compile time, you will know if you try to perform a synchronous request when you shouldn't be. Big win.

```go
resp, err := cs.AsyncRequest(&egoscale.UpdateVirtualMachineRequest{
  DisplayName: "Yoko",
}, async)
```

> cannot use req (type *egoscale.UpdateVirtualMachineRequest) as type egoscale.AsyncCommand in argument to client.AsyncRequest:
        *egoscale.UpdateVirtualMachineRequest does not implement egoscale.AsyncCommand (missing egoscale.asyncResponse method)

greut:
Being able to create a slice of requests and sending them is ❤️ 

https://github.com/exoscale/terraform-provider-exoscale/blob/new-api/src/github.com/exoscale/terraform-provider-exoscale/exoscale/compute_resource.go#L397-L402

greut:
I've renamed all the `DoFooRequest` to `DoFoo`: http://159.100.241.210/

greut:
@pyr @vincentbernat @marcaurele #roastme

greut:
All the API calls listed on our doc are done: https://community.exoscale.ch/api/compute/

greut:
manually merged.

PR #29, opened 2018-01-12
security_groups: deletion returns success as a string

greut:
...and better handling of the errors (will be used by TF)

**edit:** in fact, all the sync boolean response do so, e.g. `deleteSSHKeyPair`

> This code is needed because the response field is different for sync and async calls :(
>
> https://github.com/xanzy/go-cloudstack/blob/master/generate/generate.go#L1274


greut:
`BooleanResponse` has been renamed `booleanAsyncResponse` and be hidden away from the user. `BooleanRequest` and `BooleanAsyncRequest` return only an `error` if any.

PR #30, opened 2018-01-15
networks: createNetwork is a sync call

greut:
http://cloudstack.apache.org/api/apidocs-4.4/TOC_User.html


greut:
and a silly typo...

PR #32, opened 2018-01-15
global: make CS API command name public

greut:
This can be useful to build tools around this library (e.g. a cli)


greut:
I'll do it using `reflect` ;-)

PR #39, opened 2018-01-23
networks: name/displaytext might be empty too

greut:


greut:
@marcaurele thanks! 👏

PR #41, opened 2018-01-31
request: improve monotonic polling

greut:
## Old algorithm

Consider the polling to be instantaneous and wait for a given delay. Drawbacks, the total amount of time is unknown

https://play.golang.org/p/K8oQsB7DN1H

## New algorithm

Poll CloudStack at a fixed point in time and no more than the total amount of time given in the first place.

https://play.golang.org/p/Q3jAsk5_ORf

### DIFF
--- a/request.go
+++ b/request.go
@@ -2,6 +2,7 @@ package egoscale
 
 import (
 	"bytes"
+	"context"
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/base64"
@@ -41,6 +42,14 @@ type AsyncCommand interface {
 // `omitempty` may make sense in some cases but not all the time.
 type onBeforeHook interface {
 	onBeforeSend(params *url.Values) error
+}
+
+// AsyncInfo represents the details for any async call
+//
+// It retries at most Retries time and waits for Delay between each retry
+type AsyncInfo struct {
+	Retries int
+	Delay   int
 }
 
 const (
@@ -146,12 +155,13 @@ func (e *booleanSyncResponse) Error() error {
 	return fmt.Errorf("API error: %s", e.DisplayText)
 }
 
-// AsyncInfo represents the details for any async call
-//
-// It retries at most Retries time and waits for Delay between each retry
-type AsyncInfo struct {
-	Retries int
-	Delay   int
+type asyncJob struct {
+	command      AsyncCommand
+	delay        int
+	retries      int
+	responseChan chan<- *AsyncJobResult
+	errorChan    chan<- error
+	ctx          context.Context
 }
 
 func csQuotePlus(s string) string {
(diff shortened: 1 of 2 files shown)


greut:
Good point! I'll try this.

https://play.golang.org/p/fcYEOfrPo1O

pyr:
I think @vincentbernat know a thing or two about leaked/dangling go func which he should share.

request.go
```diff
@@ -214,64 +214,77 @@ func (exo *Client) parseResponse(resp *http.Response) (json.RawMessage, error) {
...
-	result := &QueryAsyncJobResultResponse{
-		JobStatus: job.JobStatus,
-	}
-	for async.Retries > 0 && result.JobStatus == Pending {
-		time.Sleep(time.Duration(async.Delay) * time.Second)
+		// Successful response
+		resp := req.asyncResponse()
+		if job.JobID == "" || job.JobStatus != Pending {
+			if err := json.Unmarshal(*job.JobResult, resp); err != nil {
+				channel <- err
```
vincentbernat: Missing `return`

request.go
```diff
@@ -214,64 +214,77 @@ func (exo *Client) parseResponse(resp *http.Response) (json.RawMessage, error) {
...
+			req := &QueryAsyncJobResult{JobID: job.JobID}
+			resp, err := exo.Request(req)
+			if err != nil {
+				channel <- err
+				return
+			}
+
+			result := resp.(*QueryAsyncJobResultResponse)
+			if result.JobStatus == Success {
+				channel <- result
```
vincentbernat: Missing `return`

request.go
```diff
@@ -214,64 +214,77 @@ func (exo *Client) parseResponse(resp *http.Response) (json.RawMessage, error) {
...
+		case AsyncJobResult:
+			return result, nil
+		case error:
+			return nil, result.(error)
+		default:
+			panic("Derp")
 		}
-		return errorResponse, errorResponse
+	case <-time.After(totalTime):
+		return nil, fmt.Errorf("Timeout")
```
vincentbernat: Goroutine would leak. No easy solution, everything needs to be able to be cancelled (usually with the help of a context).

vincentbernat:
My favorite live management library for Go: https://godoc.org/gopkg.in/tomb.v2 (compatible with contexts). But for a library, this may not be worth it (as you'll have to handle contexts directly as that's what a user should provide to a cancellable function).

greut:
@vincentbernat this version uses a `context` for the `Deadline`

request.go
```diff
@@ -146,12 +155,13 @@ func (e *booleanSyncResponse) Error() error {
...
-type AsyncInfo struct {
-	Retries int
-	Delay   int
+type asyncJob struct {
+	command      AsyncCommand
+	delay        int
+	retries      int
+	responseChan chan<- *AsyncJobResult
+	errorChan    chan<- error
+	ctx          context.Context
```
vincentbernat: Go style guide you mentioned a few days earlier: never embed a context into a struct.

request.go
```diff
@@ -146,12 +155,13 @@ func (e *booleanSyncResponse) Error() error {
...
-// It retries at most Retries time and waits for Delay between each retry
-type AsyncInfo struct {
-	Retries int
-	Delay   int
+type asyncJob struct {
+	command      AsyncCommand
+	delay        int
+	retries      int
+	responseChan chan<- *AsyncJobResult
+	errorChan    chan<- error
```
vincentbernat: I also like this better than runtime type check.

request.go
```diff
@@ -212,66 +222,111 @@ func (exo *Client) parseResponse(resp *http.Response) (json.RawMessage, error) {
...
 }
 
-// AsyncRequest performs an asynchronous request and polls it for retries * day [s]
-func (exo *Client) AsyncRequest(req AsyncCommand, async AsyncInfo) (interface{}, error) {
-	body, err := exo.request(req.name(), req)
+func (exo *Client) processAsyncJob(ctx context.Context, job *asyncJob) {
+	defer close(job.responseChan)
+	defer close(job.errorChan)
+
+	body, err := exo.request(job.command.name(), job.command)
```
vincentbernat: To be complete (but maybe later), you should also have a `exo.requestWithContext(ctx, ...)`.

request.go
```diff
@@ -212,66 +222,111 @@ func (exo *Client) parseResponse(resp *http.Response) (json.RawMessage, error) {
...
-	if result.JobStatus == Pending {
-		return result, fmt.Errorf("Maximum number of retries reached")
-	}
+	job.errorChan <- fmt.Errorf("Maximum number of retries reached")
+}
 
-	if err := json.Unmarshal(*result.JobResult, resp); err != nil {
-		if err := json.Unmarshal(*result.JobResult, errorResponse); err != nil {
+// AsyncRequest performs an asynchronous request and polls it for retries * day [s]
+func (exo *Client) AsyncRequest(req AsyncCommand, async AsyncInfo) (interface{}, error) {
```
vincentbernat: Same here, `AsyncRequestWithContext(ctx, ...)`

vincentbernat:
Except for the fact context needs to be propagated a bit more up and down, this looks good.

request.go
greut: I'm adding this on the todo list. #42

Issue #42, opened 2018-01-31
All calls should accept a context


greut:
**Current API**

- `client.Request(command)`
- `client.AsyncRequest(command, {Delay, Retries})`

**Additional API**

- `client.RequestWithContext(ctx, command)`
- `client.AsyncRequestWithContext(ctx, command)`, you are responsible to give a ctx with a Timeout or Cancel it.

The async request, can have a default strategy to retry on the server in a decent fashion. E.g. at x% of the remaining time (where x changes).

https://github.com/xanzy/go-cloudstack/blob/master/cloudstack/cloudstack.go#L286-L290

**Issues**

A call may take some time, and one cannot cancel it. It should accept a `context` everywhere.

> At Google, we require that Go programmers pass a Context parameter as the first argument to every function on the call path between incoming and outgoing requests. This allows Go code developed by many different teams to interoperate well. It provides simple control over timeouts and cancelation and ensures that critical values like security credentials transit Go programs properly. 
>
> https://blog.golang.org/context

cf. #41

PR #44, opened 2018-02-09
request: add context to http calls

greut:
- `timeout` parameter to the client, set by default to `60` (like `go-cloudstack`)
- `retryStrategy` parameter to the client, set by default to a Fibonacci sequence (because nerd)
- addition `RequestWithContext(ctx, command) (interface{}, error)`
- addition `AsyncRequestWithContext(ctx, command) (interface{}, error)`
- addition `BooleanRequestWithContext(ctx, command) error`
- addition `BooleanAsyncRequestWithContext(ctx, command) error`
- changes `AsyncRequest(command) (interface, error)` (no more `AsyncInfo`)

See #42


vincentbernat:
More changes than just adding context?

greut:
@vincentbernat I'm thinking about removing the `AsyncInfo` thingy. You either set a default timeout in the client if you don't care about it, or use a `Context` and manage it yourself.

vincentbernat:
I think that's a good idea (to rely on known mechanism).

greut:
@vincentbernat there is only `Command` now, and a ton of juggling with goroutines.

- `Request(Command) (interface{}, error)`
- `RequestWithContext(context.Context, Command) (interface{}, error)`
- `BooleanRequest(Command) error`
- `BooleanRequestWithContext(context.Context, Command) error`

I'll have to try it on TF to convince myself though...

greut:
:warning: breaks https://github.com/xenolf/lego/ :warning:

PR #53, opened 2018-03-13
vm: make security/affinity group ids and names mutually exclusive

greut:
`DeployVM` now requires at least one security group, either by name or id. And one cannot set both affinity group by name and id.

### DIFF
--- a/virtual_machines.go
+++ b/virtual_machines.go
@@ -1,6 +1,10 @@
 package egoscale
 
-import "net"
+import (
+	"fmt"
+	"net"
+	"net/url"
+)
 
 // VirtualMachine reprents a virtual machine
 type VirtualMachine struct {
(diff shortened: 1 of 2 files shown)


virtual_machines.go
```diff
@@ -184,6 +188,22 @@ func (*DeployVirtualMachine) APIName() string {
...
 }
 
+func (req *DeployVirtualMachine) onBeforeSend(params *url.Values) error {
+	// Either AffinityGroupIDs or AffinityGroupNames must be set
+	if len(req.AffinityGroupIDs) > 0 && len(req.AffinityGroupNames) > 0 {
+		return fmt.Errorf("Either AffinityGroupIDs or AffinityGroupNames must be set")
+	}
+
+	// Either SecurityGroupIDs or SecurityGroupNames must be set
+	if len(req.SecurityGroupIDs) == 0 && len(req.SecurityGroupNames) == 0 {
```
marcaurele: Why not leave this out? I understand that it's better to specify the SG groups but for an easier quickstart letting CS use the default SG should be enough.

greut: k, it will complain afterwards anyway.

marcaurele: How come? The VM should be created within the `default` SG group

greut: You may create a new compute resource that way.

```hcl
resource "exoscale_compute" "foo" {
    security_groups = []
}
```

Then, it'll ask you to fill the list with the actual values obtained from the API.

```console
Terraform will perform the following actions:

  ~ exoscale_compute.ada
      security_groups.#:          "1" => "0"
      security_groups.3814588639: "default" => ""

Are your really really sure? YES

* exoscale_compute.ada: A VM must have at least one Security Group, none found
```

You'd have to either update it, or remove the `security_groups` entry.

```hcl
resource "exoscale_compute" "foo" {
    security_groups = ["default"]
}
```

marcaurele: But here we're talking about the egoscale driver which should work closer to the API. On the terraform side I understand that it's a requirements to pre-fill the SG groups to avoid a change on the next apply run. Can this non-emtpy SG group can only be enforced in the terraform driver?

greut: @marcaurele the non-empty SGs rule is kind of enforced in the `UpdateVM` call as the serializer drops any empty array, thanks Go JSON serializer. It's actually a bug that turns out to be the expected behaviour.

https://github.com/exoscale/egoscale/blob/master/virtual_machines.go#L327

PR #54, opened 2018-03-15
request: add missing BooleanRequestWithContext

greut:


vincentbernat:
So, what was the problem?

greut:
@vincentbernat I was testing the same thing twice, an explict context and an one created, by the library, when you don't provide any. But wasn't testing when the `http.Client` is expected to cancel to request. In the end, one headache lead to less code, and more tests.

PR #59, opened 2018-03-20
client: generic list

greut:
In a nutshell

```go
vm := &egoscale.VirtualMachine{ZoneID: ...}
vms, err := client.List(vm)
if err != nil {
    panic(err)
}
for _, i := range vms {
    vm := i.(egoscale.VirtualMachine)
    ...
}
```

other calls:
- `ListWithContext(ctx, vm) ([]interface{}, error)`
- `AsyncList(vm) (<-chan interface{}, <-chan error)`
- `AsyncListWithContext(ctx, vm) (<-chan interface{}, <-chan error)`

**TODO**:

- [x] replace `ListNics`
- [x] replace `GetRootVolumeForVirtualMachine`

### DIFF
--- a/zones.go
+++ b/zones.go
@@ -1,6 +1,9 @@
 package egoscale
 
-import "net"
+import (
+	"context"
+	"net"
+)
 
 // Zone represents a data center
 type Zone struct {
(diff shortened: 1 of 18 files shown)


zones.go
```diff
@@ -30,6 +33,51 @@ type Zone struct {
...
+			}
+
+			zones := resp.(*ListZonesResponse)
+			for _, zone := range zones.Zone {
+				count++
+				outChan <- zone
+			}
+
+			if count >= zones.Count {
+				break
```
marcaurele: Risky! Between pagination call the count can increase or decrease. You're best bet is to stop paginating when the page result is empty.

greut: don't trust `Count` and paginate until a failure occurs?

greut: if the result size is smaller than the page size we abort.

PR #66, opened 2018-04-03
client: refactor List inner logic

greut:
`Listable.List` contains too much boilerplate, more _interface_ and _closures_ to the rescue.

```go
// current use case
vm := &egoscale.VirtualMachine{ZoneID: 1}
vms, err := client.List(vm)

// before
client.ListWithContext(ctx, vm) -> ([]interface{}, error)
    client.AsyncListWithContext(ctx, vm) -> (<-chan interface{}, <-chan error)
        vm.List(ctx, client) -> (<-chan interface{}, <-chan error)
            client.RequestWithContext(ctx, ListVirtualMachines...) -> (interface{}, error)
            ... (page 2)
            ... etc

// now
client.ListWithContext(ctx, vm)
    req <- vm.ListRequest()
    client.PaginateWithContext(ctx, req) -> ([]interface{}, error)

// new - grab the query, adapt it, run it
req, _ := vm.ListRequest()
req.Details = []string{"stats", "nics"}

client.Paginate(req, func(i interface{}, e error) bool {
    if e != nil {
        // stops the pagination
        return false
    }
    vm := i.(egoscale.VirtualMachine)
    // ...
    return true
})
```

Removed stuff:
- `AsyncList` because it wasn't working as `cancel` was called straight away
- ~~and `AsyncListWithContext` because it's unused~~

New stuff:
- `ListCommand` for all the `ListStuffs` CloudStack contains, with boilerplate `SetPage` and `SetPageSize`
- `Paginate` and `PaginateWithContext` to run your own `ListStuffs` requests, if needed.

It's a first step towards a more fluent API like `vm.DefaultNic().AddIPToNic(net.IP.Parse("..."))`

**TODO**:
- [x] put back `AsyncListWithContext` and test it
- [x] more tests using goroutines... it looks too synchronous to be ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+0.9.18
+------
+
+- feat: `VirtualMachine` is `Listable`
+- feat: new `Client.Paginate` and `Client.PaginateWithContext`
+- change: the inner logic of `Listable`
+- remove: not working `Client.AsyncList`
+
 0.9.17
 ------
 
--- a/async_jobs.go
+++ b/async_jobs.go
@@ -9,8 +9,8 @@ type AsyncJobResult struct {
 	AccountID       string           `json:"accountid"`
 	Cmd             string           `json:"cmd"`
 	Created         string           `json:"created"`
-	JobInstanceID   string           `json:"jobinstanceid"`
-	JobInstanceType string           `json:"jobinstancetype"`
+	JobInstanceID   string           `json:"jobinstanceid,omitempty"`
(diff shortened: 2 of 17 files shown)


vincentbernat:
Goroutines are leaked when pagination is not used until the end. See https://www.jtolio.com/2016/03/go-channels-are-bad-and-you-should-feel-bad/. But maybe we can argue that context enables to kill the goroutine after its use (people should be told to never just drop `context.Background()` then).

greut:
screenshot from 2018-04-09 08-35-48

PR #76, opened 2018-04-12
replace ListCommand implementations with pointers

### LINKED ISSUE #75: Replace IterateItemFunc to use pointers [rfc]
When use Paginate, the callback function receives the value itself rather than a pointer to it. It's premature optimization but it would be more efficient and similar the the other calls to have a pointer instead.

```go
// direct call - no pagination
r, _ := client.Request(&egoscale.ListAPIs{})
apis := r.(*egoscale.ListAPIsResponse).API

// whole list with Pagination
apis, _ := client.List(&egoscale.API{})
for _, i := range apis {
    // now
    api := i.(egoscale.API)
    // proposal
    api := i.(*egoscale.API)

    // ...
}

// iterate with pagination
client.Paginate(&egoscale.ListAPIs, func(i interface{}, e error) bool {
    // now
    api := i.(egoscale.API)
    // proposal
    api := i.(*egoscale.API)

    // ...
})
```

greut:
closes #75 

**TODO**
- [x] ListNics
- [x] ListSSHKeyPairs
- [x] ListTemplates
- [x] ListVirtualMachines
- [x] ListVolumes
- [x] ListZones

### DIFF
--- a/addresses.go
+++ b/addresses.go
@@ -94,3 +94,44 @@ func (*ListPublicIPAddresses) APIName() string {
 func (*ListPublicIPAddresses) response() interface{} {
 	return new(ListPublicIPAddressesResponse)
 }
+
+// ListRequest builds the ListAdresses request
+func (ipaddress *IPAddress) ListRequest() (ListCommand, error) {
+	req := &ListPublicIPAddresses{
+		Account:            ipaddress.Account,
+		AllocatedNetworkID: ipaddress.AssociatedNetworkID,
+		DomainID:           ipaddress.DomainID,
+		ForDisplay:         &ipaddress.ForDisplay,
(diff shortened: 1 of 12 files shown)


addresses.go
```diff
@@ -94,3 +94,44 @@ func (*ListPublicIPAddresses) APIName() string {
 func (*ListPublicIPAddresses) response() interface{} {
 	return new(ListPublicIPAddressesResponse)
 }
+
+// ListRequest builds the ListAdresses request
+func (ip *IPAddress) ListRequest() (ListCommand, error) {
```
greut: rename it `ipaddress *IPAddress` (see codeclimate) and I'd move it next to the other `IPAddress` methods.

greut:
@pierre-emmanuelJ thanks!

addresses.go
pierre-emmanuelJ: 👍

PR #77, opened 2018-04-13
listcommand: check type assertions

greut:
type assertions are good... we shall use them.

PS: it is expected that the code coverage goes down...


pierre-emmanuelJ:
It's good way to throw up errors ! 👍

PR #78, opened 2018-04-13
volumes: implements gettable

greut:


volumes.go
```diff
@@ -36,6 +41,29 @@ func (*Volume) ResourceType() string {
 	return "Volume"
 }
```
greut: !godoc

volumes.go
```diff
@@ -36,6 +41,29 @@ func (*Volume) ResourceType() string {
...
 
+func (vol *Volume) Get(ctx context.Context, client *Client) error {
```
pierre-emmanuelJ: Missing comment

PR #85, opened 2018-04-17
snapshots: fill up the documentation

greut:
`generate` found some missing fields, big win!

- `createSnapshot.name`
- `listSnapshots.ids`


pierre-emmanuelJ:
👍

PR #91, opened 2018-04-19
request: generate stringer from enums

greut:
for the pleasure of using `go generate`


pierre-emmanuelJ:
better than make stringer cmd manually

PR #104, opened 2018-04-26
request: send user-agent

greut:
Sends:
- request as GET by default
- User-Agent containing exoscale/egoscale and a version number


dns.go
```diff
@@ -219,6 +219,7 @@ func (exo *Client) dnsRequest(uri string, params string, method string) (json.Ra
 
 	var hdr = make(http.Header)
 	hdr.Add("X-DNS-TOKEN", exo.APIKey+":"+exo.apiSecret)
+	hdr.Add("User-Agent", fmt.Sprintf("exoscale/egoscale (%v)", Version))
```
marcaurele: 🤠

PR #121, opened 2018-05-02
snapshots: fix docs

greut:


snapshots_type.go
```diff
@@ -31,7 +31,6 @@ type Snapshot struct {
 type CreateSnapshot struct {
 	VolumeID  string `json:"volumeid" doc:"The ID of the disk volume"`
 	Account   string `json:"account,omitempty" doc:"The account of the snapshot. The account parameter must be used with the domainId parameter."`
-	Name      string `json:"name,omitempty" doc:"the name of the snapshot"`
```
marcaurele: Don't remove that one, it's not used today but should be

greut: oh ok

PR #126, opened 2018-05-03
vm: add migrateVirtualMachine

greut:

### DIFF
--- a/virtual_machines_type.go
+++ b/virtual_machines_type.go
@@ -345,3 +345,15 @@ type GetVirtualMachineUserDataResponse struct {
 	UserData         string `json:"userdata,omitempty" doc:"Base 64 encoded VM user data"`
 	VirtualMachineID string `json:"virtualmachineid,omitempty" doc:"the ID of the virtual machine"`
 }
+
+// MigrateVirtualMachine (Async) attempts migration of a VM to a different host or Root volume of the vm to a different storage pool
+//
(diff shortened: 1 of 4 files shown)


virtual_machines_type.go
```diff
@@ -343,3 +343,15 @@ type GetVirtualMachineUserDataResponse struct {
 	UserData         string `json:"userdata,omitempty" doc:"Base 64 encoded VM user data"`
 	VirtualMachineID string `json:"virtualmachineid,omitempty" doc:"the ID of the virtual machine"`
 }
+
+// MigrateVirtualMachine (Async) attempts migration of a VM to a different host or Root volume of the vm to a different storage pool
+//
+// CloudStack API: http://cloudstack.apache.org/api/apidocs-4.10/apis/migrateVirtualMachine.html
+type MigrateVirtualMachine struct {
+	HostID           string `json:"hostid,omitempty" doc:"Destination Host ID to migrate VM to. Required for live migrating a VM from host to host"`
```
marcaurele: In our setup, we never call this `migrateVirtualMachine` with a host id but only with the storagepool id. To live migrate we need to migrate both the VM and the Volume together which is done with `migrateVirtualMachineWithVolume` instead.

The host id is used when the VM is using remote disk volumes, which is not our setup with the local root volume.

greut: okay. (all of this comes from listApis)

PR #132, opened 2018-05-07
Less types

greut:
An attempt to remove most of the `<Foo>Response` type. We have to keep them for the `List<Foo>s` commands though.

**Old**

https://godoc.org/github.com/exoscale/egoscale#pkg-index

**New**

screenshot-2018-5-7 egoscale - the go programming language

PR #134, opened 2018-05-14
snapshot: add listSnapshotPolicies

greut:
It'll be easier to test with #132


greut:
we don't do snapshot policies, #129

PR #145, opened 2018-05-28
gorelease: use skip-sign within travis

greut:


greut:
but travis stopped running the tests??

PR #154, opened 2018-06-04
Add test on body response

pierre-emmanuelJ:

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -540,12 +540,9 @@ func TestWrongBodyResponse(t *testing.T) {
 	`})
 	defer ts.Close()
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
-	defer cancel()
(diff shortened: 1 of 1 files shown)


request_test.go
```diff
@@ -529,6 +529,32 @@ func TestBooleanRequestWithContextAndTimeout(t *testing.T) {
...
+		</html>		
+	`})
+	defer ts.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+
+	cs := NewClient(ts.URL, "TOKEN", "SECRET")
+
+	_, err := cs.RequestWithContext(ctx, &ListZones{})
```
greut: you may drop the `ctx` as that's not the tested thing here.

pierre-emmanuelJ: yep

PR #168, opened 2018-06-13
add print cmd

pierre-emmanuelJ:

### DIFF
--- a/cmd/exo/cmd/ssh.go
+++ b/cmd/exo/cmd/ssh.go
@@ -29,10 +29,21 @@ func sshCmdRun(cmd *cobra.Command, args []string) {
 		log.Fatal(err)
 	}
 
+	isConnectionSTR, err := cmd.Flags().GetBool("print")
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	infos, err := getSSHInfos(args[0])
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	if isConnectionSTR {
+		printSSHConnectSTR(infos)
+		return
+	}
+
 	if isInfos {
 		printSSHInfos(infos)
 		return
@@ -75,6 +86,10 @@ func getSSHInfos(name string) (*sshInfos, error) {
 
 }
 
+func printSSHConnectSTR(infos *sshInfos) {
+	fmt.Printf("ssh -i %s %s@%s\n", infos.sshKeys, infos.userName, infos.ip)
+}
+
 func printSSHInfos(infos *sshInfos) {
 	println("Virtual machine name", infos.vmName, "with ID", infos.vmID)
 	println(" - sshkey path:", infos.sshKeys)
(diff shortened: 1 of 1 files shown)


pierre-emmanuelJ:
It can be cool with --ip6 flag 👍

pierre-emmanuelJ:
@greut

greut:
@pierre-emmanuelJ

cmd/exo/cmd/ssh.go
```diff
@@ -48,12 +64,27 @@ type sshInfos struct {
...
 
+	nic := vm.DefaultNic()
+	if nic == nil {
+		return nil, fmt.Errorf("No default NIC on this instance")
+	}
+
+	vmIP := vm.IP()
+
+	if isIpv6 {
+		if nic.IP6Address.String() != "<nil>" {
```
greut: `nic.IP6Address != nil`

cmd/exo/cmd/ssh.go
```diff
@@ -68,13 +99,17 @@ func getSSHInfos(name string) (*sshInfos, error) {
...
-		ip:       vm.IP(),
+		ip:       vmIP,
 		vmName:   vm.Name,
 		vmID:     vm.ID,
 	}, nil
 
 }
 
+func printSSHConnectSTR(infos *sshInfos) {
+	fmt.Printf("ssh -i %s %s@%s\n", infos.sshKeys, infos.userName, infos.ip)
```
greut: I'd love to see a check for the presence of the ssh key file.

pierre-emmanuelJ: i had think about that this morning too. but forgotten to write it somewhere

cmd/exo/cmd/ssh.go
```diff
@@ -70,6 +70,12 @@ func getSSHInfos(name string, isIpv6 bool) (*sshInfos, error) {
 		return nil, err
 	}
 
+	sshKeyPath := path.Join(configFolder, "instances", vm.ID, "id_rsa")
+
+	if _, err := os.Stat(sshKeyPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("SSH keypair: %q not found", sshKeyPath)
+	}
```
greut: good enough for now.

pierre-emmanuelJ: yes

greut:
With the SSH feature, we have to use cases:
1. you want to connect to an existing instance
2. you want to connect to an instance created with `exo`

In the first use case, the ssh key to be used likely lives in the _ssh agent_ configuration, hence we should ignore the `~/exoscale/instance/...` if it doesn't exist at all instead of breaking.

cmd/exo/cmd/ssh.go
pierre-emmanuelJ: in this case `nic.IP6Address` is not `nil`

greut: then the opposite...

PR #169, opened 2018-06-14
Bulma

greut:
A nicer, bigger README... without the red.

screenshot from 2018-06-14 11-48-16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ jobs:
       cd ../..
       goreleaser --snapshot --skip-sign
   - stage: gh-pages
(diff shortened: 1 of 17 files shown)


pierre-emmanuelJ:
👍 Nice

greut:
screenshot from 2018-06-14 15-44-15
screenshot from 2018-06-14 15-43-55

greut:
>  As mentioned previously, if it's feasible I'd love for the cs functionality to be in exo directly (exo api-call ... for instance) instead of a separate tool.

It feels like a lot a work for no added value.

PR #170, opened 2018-06-15
Remove double call to config when first run

pierre-emmanuelJ:

### DIFF
--- a/cmd/exo/cmd/root.go
+++ b/cmd/exo/cmd/root.go
@@ -120,7 +120,7 @@ func initConfig() {
 	}
 }
 
-// return a command position in by fetching os.args and ignoring flags
+// return a command position by fetching os.args and ignoring flags
 //
 //example: "$ exo -r preprod vm create" vm position is 1 and create is 2
 //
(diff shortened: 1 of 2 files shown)


pierre-emmanuelJ:
Problem:

If no config file exist `exo` try to build config file by calling `generateConfig()`

so if you call `exo config` if no config file exist it will call config two time.

greut:
how can I reproduce the issue?

pierre-emmanuelJ:
Not on this commit:

Remove all your config file, then `$ exo config`

greut:
`config` needs a lot of love.

cmd/exo/cmd/root.go
```diff
@@ -104,15 +105,56 @@ func initConfig() {
...
-			log.Fatal(err)
-		}
-		configFilePath = path
+	if cfgFilePath != "" {
+		configFilePath = cfgFilePath
+		return
+	}
+
+	if getCmdPosition("config") == 1 {
+		ignoreBuilClient = true
```
greut: typo?

pierre-emmanuelJ: `ignoreClientBuild` instead of `ignoreBuilClient` and missing `d` to build ! thanks 👍

PR #147, opened 2018-05-29
DNS cmd

pierre-emmanuelJ:
Signed-off-by: Pierre-Emmanuel Jacquier <pierre-emmanuel.jacquier@epitech.eu>

PR #184, opened 2018-06-26
dns: backporting GetDomains

greut:
`GetDomains` will be needed for the terraform provider, hence I've stolen @pierre-emmanuelJ code from #147


dns.go
```diff
@@ -64,6 +82,41 @@ type DNSError struct {
 	Name []string `json:"name"`
 }
 
+// Record represent record type
+type Record int
+
+//go:generate stringer -type=Record
```
pierre-emmanuelJ: Where is this stringers generated file ?

record_string.go
```diff
@@ -0,0 +1,16 @@
+// Code generated by "stringer -type=Record"; DO NOT EDIT.
```
greut: @pierre-emmanuelJ here?

pierre-emmanuelJ: ohww my bad !

PR #190, opened 2018-06-28
exo: add auto-completion

pierre-emmanuelJ:
- [x] bash


greut:
no changes in `.travis` or `goreleaser`?

pierre-emmanuelJ:
WIP, i will add after some tests

.travis.yml
```diff
@@ -46,6 +46,8 @@ jobs:
       dep ensure -v -vendor-only
       cd ../..
       goreleaser --snapshot --skip-sign
+      cd cmd/exo
+      go run ./completion/main.go
```
greut: the goal would be to include those files in the release, or do I miss something?

pierre-emmanuelJ: yes

greut:
for Bash it seems ok. Do you know why the zsh file contains so little?

```console
% exo 
affinitygroup  eip            privnet        sshkey         vm             
config         firewall       ssh            template       zone
```

pierre-emmanuelJ:
i dont know but i think we will be forced to use bash_completion to make zsh completion like cs

.travis.yml
```diff
@@ -45,6 +45,10 @@ jobs:
       cd ../exo
       dep ensure -v -vendor-only
```
greut:
```
go run bash_completion/main.go
mv bash_completion ../..
```

.travis.yml
```diff
@@ -45,6 +45,10 @@ jobs:
...
       cd ../..
+      cd cmd/exo/completion
+      go run main.go
+      cp bash_completion ../../../
```
greut: or `go run cmd/exo/completion`

pierre-emmanuelJ: 👍

.travis.yml
pierre-emmanuelJ: 👇

PR #194, opened 2018-06-29
global: remove hackish makefiles

greut:
make things more familiar to newcomers.

```
dep ensure
go build
go install
```


pierre-emmanuelJ:
why you remove it ? because goreleaser ?

greut:
it's fugly and we should either add one in `exo` or remove them all. Removing code feels nicer.

PR #203, opened 2018-07-02
addresses: list cannot be that magic

greut:
Since the values don't have `*`, `ListRequest` should take into account what is the default value on the cloudstack side. This fix should probably be adapted to other types as well.


pierre-emmanuelJ:
If you need help I can looking for the other types to fix it

greut:
let's hope I got them all.

PR #212, opened 2018-07-03
implem privnet associate, dissociate, show

pierre-emmanuelJ:

### DIFF
--- a/cmd/exo/cmd/privnet.go
+++ b/cmd/exo/cmd/privnet.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"log"
-	"strings"
 
 	"github.com/exoscale/egoscale"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


pierre-emmanuelJ:
missing just `show` and `list` vm details

cmd/exo/cmd/privnet.go
```diff
@@ -15,28 +15,28 @@ var privnetCmd = &cobra.Command{
...
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	res := ""
+	var res *egoscale.Network
 	match := 0
 	for _, net := range nets {
 		n := net.(*egoscale.Network)
 		if strings.Compare(name, n.Name) == 0 || strings.Compare(name, n.ID) == 0 {
```
greut: :book:

> Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on. 
>
> https://godoc.org/strings#Compare

cmd/exo/cmd/privnet_show.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		table.Render()
+
+		return nil
+	},
+}
+
+func privnetDetails(privnetName string) (*egoscale.Network, []egoscale.VirtualMachine, error) {
+
+	resp, err := cs.List(&egoscale.Zone{})
```
greut: s/resp/zones/ ?

cmd/exo/cmd/privnet_show.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	network, err := searchPrivnet(privnetName, resp)
+	if err != nil {
+		return nil, nil, err
+	}
```
greut: I don't get why you're doing O(n) calls instead of O(1) here? It's a bit more work to filter the networks and create a `map[zone]network` but it should be worth it.

I would inform the user if more than one network was found for a given name for any show/delete operation

PR #226, opened 2018-07-13
exo: read more env variables

greut:
to have the same ones as docker-machine

https://docs.docker.com/machine/drivers/exoscale/#environment-variables-and-default-values


marcaurele:
Can you put a comment too in the code to say this is related to docker-machine environment variables, thanks

greut:
then, I'll add all the ones from CS :-)

https://github.com/exoscale/terraform-provider-exoscale/blob/master/website/docs/index.html.markdown#environment-variables

greut:
it's not even consistent with what is in docker machine... :man_facepalming:

cmd/exo/cmd/root.go
```diff
@@ -215,3 +233,13 @@ func getCmdPosition(cmd string) int {
 
 	return count
 }
+
+// readFromEnv is a os.Getenv on steroids
+func readFromEnv(keys ...string) string {
```
marcaurele: `crystalBall(keys ...string) string {` ...  🙃

PR #233, opened 2018-07-18
Sync request

greut:
this exposes two methods:
- `SyncRequest`
- `SyncRequestWithContext`

because there was not way to perform a fire and forget on an asynchronous command.

Another change.

```
// was
err := jobResult.Response(virtualMachine)
// now
err := jobResult.Result(virtualMachine)
```

But, I'm unhappy with it. Should it be called `jobResult.Unmarshal(virtualMachine)`, `Load`, `Hydrate` ?

### DIFF
--- a/request.go
+++ b/request.go
@@ -222,6 +222,37 @@ func (client *Client) RequestWithContext(ctx context.Context, request Command) (
 	}
 }
 
+// SyncRequest performs the async command in a sync way (job id)
+func (client *Client) SyncRequest(request AsyncCommand) (*AsyncJobResult, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), client.Timeout)
+	defer cancel()
+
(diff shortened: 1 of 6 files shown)


request.go
```diff
@@ -117,34 +117,38 @@ func (client *Client) parseResponse(resp *http.Response, key string) (json.RawMe
...
 	var err error
 
-	res := request.asyncResponse()
-	client.AsyncRequestWithContext(ctx, request, func(j *AsyncJobResult, er error) bool {
-		if er != nil {
-			err = er
-			return false
-		}
-		if j.JobStatus == Success {
-			if r := j.Response(res); err != nil {
```
greut: @pierre-emmanuelJ here the `err != nil` was probably wrong.

greut:
w00t, code coverage is back :-)

PR #239, opened 2018-07-21
nic: use MACAddress rather than string

greut:
Making it even more opinionated.

@vincentbernat embed struct vs type aliases are probably equivalent in terms of boilerplate work around them.


macaddress.go
```diff
@@ -0,0 +1,55 @@
...
+	m[2] = c
+	m[3] = d
+	m[4] = e
+	m[5] = f
+	return MACAddress{m}
+}
+
+// UnmarshalJSON unmarshals the raw JSON into the MAC address
+func (mac *MACAddress) UnmarshalJSON(b []byte) error {
+	addr := strings.Trim(string(b), "\"")
```
vincentbernat: That makes `00:11:22:33:44:5"` (encoded as `"00:11:22:33:44:5\""`) a valid MAC address. Safer to just use `json.Unmarshal(&macString)` with `var macString string`.

greut:
```
--- FAIL: TestListNics (0.00s)
        nics_test.go:56: address 06:f0:f8:00:00:5\: invalid MAC address
```

greut:
```diff
--- a/nics_test.go
+++ b/nics_test.go
@@ -37,7 +37,7 @@ func TestListNics(t *testing.T) {
                        "id": "fed8fa77-c27d-411c-a630-2ce787161ad6",
                        "ipaddress": "165.150.8.20",
                        "isdefault": true,
-                       "macaddress": "06:f0:f8:00:00:57",
+                       "macaddress": "06:f0:f8:00:00:5\"",
```

greut:
the only win was `String()` (https://godoc.org/net#HardwareAddr) while losing the convenient casting.

PR #240, opened 2018-07-23
dep: bundle jinzhu/copier in vendor

greut:
as jinzhu/copier is one file, let's vendorize it for good.


vincentbernat:
Why? Also, vendoring doesn't work for libraries, right?

greut:
to not force `dep` on the rest of the world.

PR #261, opened 2018-08-03
implem exo vm stop

pierre-emmanuelJ:

### DIFF
--- a/cmd/exo/cmd/vm.go
+++ b/cmd/exo/cmd/vm.go
@@ -12,7 +12,6 @@ var vmCmd = &cobra.Command{
 }
 
 func getVMWithNameOrID(cs *egoscale.Client, name string) (*egoscale.VirtualMachine, error) {
-
 	vm := &egoscale.VirtualMachine{ID: name}
 	if err := cs.Get(vm); err == nil {
 		return vm, err
--- a/cmd/exo/cmd/vm.go
+++ b/cmd/exo/cmd/vm.go
@@ -11,7 +11,7 @@ var vmCmd = &cobra.Command{
 	Short: "Virtual machines management",
 }
 
-func getVMWithNameOrID(cs *egoscale.Client, name string) (*egoscale.VirtualMachine, error) {
+func getVMWithNameOrID(name string) (*egoscale.VirtualMachine, error) {
 	vm := &egoscale.VirtualMachine{ID: name}
 	if err := cs.Get(vm); err == nil {
 		return vm, err
--- a/cmd/exo/cmd/vm_stop.go
+++ b/cmd/exo/cmd/vm_stop.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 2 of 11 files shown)


cmd/exo/cmd/vm.go
```diff
@@ -11,11 +11,23 @@ var vmCmd = &cobra.Command{
...
 }
 
-func getVMWithNameOrID(cs *egoscale.Client, name string) (*egoscale.VirtualMachine, error) {
+// nolint
+const (
+	starting = "Starting"
+	running  = "Running"
+	stoping  = "Stopping"
+	stopped  = "Stopped"
+)
```
greut: Is it worth creating it in egoscale then?

https://github.com/apache/cloudstack/blob/893a88d225276e45f12f9490e6af2c94a81c2965/api/src/main/java/com/cloud/vm/VirtualMachine.java#L46-L55

pierre-emmanuelJ: I put it in egoscale is better

virtual_machines.go
```diff
@@ -11,6 +11,20 @@ import (
 	"net/url"
 )
 
+// nolint
```
greut: it's called cheating

cmd/exo/cmd/vm_stop.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		for _, v := range args {
+			if err := asyncStopWithCtx(gContext, v); err != nil {
+				fmt.Fprintln(os.Stderr, err) // nolint: errcheck
+			} else {
+				fmt.Println("\nStopped !")
+			}
+		}
+		return nil
```
greut: I don't like this `return nil`

pierre-emmanuelJ: I dont want to stop all args vm stopping so I just display error on stderr and continue the flow.
that why I return nil at the end

greut: gather all the errors and return one of them in case of a failure.

cmd/exo/cmd/vm_stop.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			errorReq = err
+			return false
+		}
+
+		if jobResult.JobStatus == egoscale.Success {
+			return false
+		}
+		return true
+	})
```
greut: do the final `Println` here.

greut:
```console
% exo vm stop foo bar
could not stop "foo": virtual machine is not running
could not stop "bar": API error ParamError 431 (CSErrorCode(0) 0): not found, query: apikey=EXOc94bc655ff901b7218a5c3cd, command=listVirtualMachines, name=bar, response=json

% exo vm stop foo bar
Stopping "foo" .. success.
could not stop "bar": API error ParamError 431 (CSErrorCode(0) 0): not found, query: apikey=EXOc94bc655ff901b7218a5c3cd, command=listVirtualMachines, name=bar, response=json

% echo $?
1
```

pierre-emmanuelJ:
okay

cmd/exo/cmd/vm_stop.go
```diff
@@ -0,0 +1,84 @@
...
+			return errors.New(b.String())
+		}
+
+		return nil
+	},
+}
+
+// stopVirtualMachine stop a virtual machine instance
+func stopVirtualMachine(vmName string) error {
+	vm, err := getVMWithNameOrID(cs, vmName)
```
greut: `cs` is a global variable. We don't have to repeat it here.

PR #269, opened 2018-08-06
rename Force into Must to follow regexp

greut:
be consistent... https://golang.org/pkg/regexp/#MustCompile


pierre-emmanuelJ:
👍

Issue #268, opened 2018-08-06
[RFC] force UUID rather than string


greut:
Many identifiers (`ID`) are in fact UUIDv4 which some can be turned in long. But as this is expected to disappear for obvious reasons, this library should reflect it with a stronger type for all the `ID` fields.

greut:
we'll do it for 0.11

PR #271, opened 2018-08-06
zone: use UUID instead of string

greut:
cf #268 

```console
% cs listZones --id=1
Incorrect Usage: invalid value "1" for flag -id: not a valid UUID, got 1
```


pierre-emmanuelJ:
👍

PR #272, opened 2018-08-06
Implem exo vm reset (restore)

pierre-emmanuelJ:

### DIFF
--- a/cmd/exo/cmd/vm_reset.go
+++ b/cmd/exo/cmd/vm_reset.go
@@ -9,8 +9,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var resetCmdDiskSize *int64
-
 // vmResetCmd represents the stop command
 var vmResetCmd = &cobra.Command{
 	Use:   "reset <vm name> [vm name] ...",
@@ -25,9 +23,14 @@ var vmResetCmd = &cobra.Command{
 			return err
 		}
 
+		force, err := cmd.Flags().GetBool("force")
+		if err != nil {
+			return err
+		}
+
 		errs := []error{}
 		for _, v := range args {
-			if err := resetVirtualMachine(v, diskValue); err != nil {
+			if err := resetVirtualMachine(v, diskValue, force); err != nil {
 				errs = append(errs, fmt.Errorf("could not reset %q: %s", v, err))
 			}
 		}
@@ -50,10 +53,16 @@ var vmResetCmd = &cobra.Command{
 }
 
 // resetVirtualMachine stop a virtual machine instance
-func resetVirtualMachine(vmName string, diskValue int64PtrValue) error {
+func resetVirtualMachine(vmName string, diskValue int64PtrValue, force bool) error {
 	vm, err := getVMWithNameOrID(vmName)
 	if err != nil {
 		return err
+	}
+
+	if !force {
+		if !askQuestion(fmt.Sprintf("sure you want to reset %q virtual machine", vm.Name)) {
+			return nil
+		}
 	}
 
 	volume := &egoscale.Volume{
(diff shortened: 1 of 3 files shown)


greut:
I guess, you'll have to send at least the `rootdisksize` option. @marcaurele do you confirm?

greut:
like for the deletion, I'd ask the user confirmation before restoring everything from scratch :stuck_out_tongue_winking_eye:

marcaurele:
Yes it's kind of mandatory to send the `rootdisksize` otherwise the template size will be used, which is now not the correct value due to the single template size at 10GB. For the reset operation the user can downscale its disk. It would be nice to present the current disk size and let it choose any other value in the usual boundaries: starting at the template size of the one selected, maxed by the SO of the VM.

cmd/exo/cmd/vm_reset.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			return errors.New(b.String())
+		}
+
+		return nil
+	},
+}
+
+// resetVirtualMachine stop a virtual machine instance
+func resetVirtualMachine(vmName string, diskValue int64PtrValue, force bool) error {
```
greut: you should reapply the current disk size.

```diff
- func resetVirtualMachine(vmName string, diskValue int64PtrValue, force bool) error {
+ func resetVirtualMachine(vmName string, force bool) error {
```

request.go
```diff
@@ -361,11 +361,14 @@ func (client *Client) request(ctx context.Context, command Command) (json.RawMes
...
+	// XXX: addIpToNic, activateIp6, restorevmresponse are kind of special
+	switch key {
+	case "addiptonicresponse":
 		key = "addiptovmnicresponse"
-	} else if key == "activateip6response" {
+	case "activateip6response":
 		key = "activateip6nicresponse"
+	case "restorevirtualmachineresponse":
+		key = "restorevmresponse"
 	}
```
greut: maybe, we could have a map or but that into the client, like APIName/APIDescription

cmd/exo/cmd/vm_reset.go
greut: my bad, you did well.

request.go
pierre-emmanuelJ: 👍

cmd/exo/cmd/vm_reset.go
```diff
@@ -74,11 +74,21 @@ func resetVirtualMachine(vmName string, diskValue int64PtrValue, force bool) err
 		return err
 	}
 
-	rootDiskSize := int64(volume.Size >> 30)
+	volumeSize := volume.Size >> 30
+
+	temp := &egoscale.Template{IsFeatured: true, ID: vm.TemplateID, ZoneID: "1"}
```
greut: reuse an existing ZoneID (cf #271)

cmd/exo/cmd/vm_reset.go
```diff
@@ -0,0 +1,127 @@
...
+		Type:             "ROOT",
+	}
+
+	if err := cs.Get(volume); err != nil {
+		return err
+	}
+
+	volumeSize := volume.Size >> 30
+
+	temp := &egoscale.Template{IsFeatured: true, ID: vm.TemplateID, ZoneID: "1128bd56-b4d9-4ac6-a7b9-c715b187ce11"}
```
greut: `vm.ZoneID` ?

request.go
```diff
@@ -361,11 +361,10 @@ func (client *Client) request(ctx context.Context, command Command) (json.RawMes
 
 	apiName := client.APIName(command)
 	key := fmt.Sprintf("%sresponse", strings.ToLower(apiName))
-	// XXX: addIpToNic, activateIp6 are kind of special
-	if key == "addiptonicresponse" {
-		key = "addiptovmnicresponse"
-	} else if key == "activateip6response" {
-		key = "activateip6nicresponse"
+	// XXX: addIpToNic, activateIp6, restorevmresponse are kind of special
```
greut: feel free to move the comment closer to the map.

cmd/exo/cmd/vm_reset.go
pierre-emmanuelJ: yes

PR #275, opened 2018-08-07
exo: async request

pierre-emmanuelJ:


cmd/exo/cmd/request.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"fmt"
+
+	"github.com/exoscale/egoscale"
+)
+
+// asyncRequest if no response expected send nil
+func asyncRequest(cmd egoscale.AsyncCommand, msg string, response interface{}) error {
+	fmt.Print(msg)
+	var errorReq error
```
greut:
```golang
response := cs.Response(cmd)
cs.AsyncRequestWithContext(gContext, cmd, ...)
return response, errorReq
```

cmd/exo/cmd/request.go
```diff
@@ -0,0 +1,40 @@
...
+			return false
+		}
+
+		if jobResult.JobStatus == egoscale.Success {
+			if response != nil {
+				if errR := jobResult.Result(response); errR != nil {
+					errorReq = errR
+				}
+			}
+			fmt.Println(" success.")
```
greut: how come do I get a `success` when the _umarshaling_ failed?

cmd/exo/cmd/request.go
pierre-emmanuelJ: I can't return an empty interface because I can't create it in the function.
So It's better to give in parameter and more simple

greut: This should work.

```diff
-func asyncRequest(cmd egoscale.AsyncCommand, msg string, response interface{}) error {
+func asyncRequest(cmd egoscale.AsyncCommand, msg string) (interface{}, error) {
+    response := cs.Response(cmd)
     ...
-    return errReq
+    return response, errReq
```

PR #277, opened 2018-08-07
client: unbreak

greut:
“Just keep in mind that if any concrete value has been stored in the
 interface, the interface will not be nil”
                          -- https://golang.org/doc/faq#nil_error


pierre-emmanuelJ:
👍

PR #281, opened 2018-08-09
Point to cmd installation instructions from README

chrisglass:
While installing the CLI, I noticed a couple of small touch-ups that could be made to the READMEs

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ description: the Go library for Exoscale
 
 A wrapper for the [Exoscale public cloud](https://www.exoscale.com) API.
 
+This project also provides an executable CLI client. See [the relevant subtree](https://github.com/exoscale/egoscale/tree/master/cmd/exo) for installation instructions.
+
 ## License
 
 Licensed under the Apache License, Version 2.0 (the "License"); you
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ description: the Go library for Exoscale
 
 A wrapper for the [Exoscale public cloud](https://www.exoscale.com) API.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,6 +9,8 @@ description: the Go library for Exoscale
 
 A wrapper for the [Exoscale public cloud](https://www.exoscale.com) API.
 
+This project also provides an executable CLI client. See [the relevant subtree](https://github.com/exoscale/egoscale/tree/master/cmd/exo) for installation instructions.
+
```
pierre-emmanuelJ: That a good point 👍

cmd/exo/README.md
```diff
@@ -25,7 +25,12 @@ $ go install
 ## Configuration
 
 A configuration file holding your credentials is required.
-You can generate one via a guided prompt.
+You can generate one via a guided interactive prompt with:
+
+```
+$ exo config
+```
+
```
greut: as it duplicates what's below under _automatic configuration_, why not removing the _You can generate..._ line?

README.md
chrisglass: Ah yeah, even better!

:+1:

cmd/exo/README.md
chrisglass: Fair enough :)

chrisglass: Actually, how about moving the "Automatic configuration" header to be "configuration"?

/me does

greut:
the `exo config` output is outdated, feel free to remove it as well.

greut:
thank you very much @chrisglass

PR #297, opened 2018-08-20
Fix handling of brackets '[]'

stffabi:
Fix handling of brackets '[]' in parameter values (e.g.in a description field) and 
when a parameter name contains
'[]' and a POST is used (e.g. during a DeployVirtualMachine).

CloudStack only uses a specialized encoding for the signature
calculation. It only query encodes the values of the query arguments
and always use '%20' for whitespaces (no '+'). But for the GET url
and for the POST body, we always have to use the normal
query escaping as standardized.

Signed-off-by: stffabi <stffabi@pm.me>

### DIFF
--- a/client.go
+++ b/client.go
@@ -117,12 +117,13 @@ func (client *Client) GetWithContext(ctx context.Context, g Gettable) error {
 		if err != nil {
 			return err
 		}
-		payload, err := client.Payload(req)
+		params, err := client.Payload(req)
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


greut:
Thanks for the fix and the appreciated refactoring. It check this!

greut:
It's much nicer indeed. thanks again for your contributions @stffabi

stffabi:
You're welcome, thanks for merging.

PR #304, opened 2018-08-30
request: signatureVersion=3

greut:
Enables the API call expiration, 10 minutes by default.

http://docs.cloudstack.apache.org/en/4.11.1.0/developersguide/dev.html?enabling-api-call-expiration

### DIFF
--- a/request.go
+++ b/request.go
@@ -297,7 +297,7 @@ func (client *Client) Payload(command Command) (url.Values, error) {
 	params.Set("command", client.APIName(command))
 	params.Set("response", "json")
 	params.Set("signatureversion", "3")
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -18,6 +18,9 @@ import (
 	"time"
 )
 
+// requestValidity represents when a signed request is meant to expire
+const requestValidity = 10 * time.Second
```
marcaurele: Too short, it should handle some gap for desynchronized clock. What about 10 minutes?

marcaurele: We can also have this as a parameter with a default value so that a configuration file for a tool can set this value to something different.

greut: if you put it yourself, _egoscale_ won't override it.

greut:
cf https://github.com/exoscale/cs/pull/70

PR #309, opened 2018-09-04
client: higher code coverage (even fake)

greut:
let's kick this week off :)

### DIFF
--- a/client.go
+++ b/client.go
@@ -308,7 +308,7 @@ func (client *Client) APIName(command Command) string {
 func (client *Client) APIDescription(command Command) string {
 	info, err := info(command)
 	if err != nil {
-		panic(err)
+		return "*missing description*"
 	}
 	return info.Description
 }
--- a/client_test.go
+++ b/client_test.go
@@ -21,6 +21,14 @@ func TestClientAPIName(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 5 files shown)


client.go
```diff
@@ -308,7 +308,7 @@ func (client *Client) APIName(command Command) string {
 func (client *Client) APIDescription(command Command) string {
 	info, err := info(command)
 	if err != nil {
-		panic(err)
+		return "*missing description*"
```
pierre-emmanuelJ: description is comming

client_test.go
```diff
@@ -21,6 +21,14 @@ func TestClientAPIName(t *testing.T) {
...
 }
 
+func TestClientAPIDescription(t *testing.T) {
+	cs := NewClient("ENDPOINT", "KEY", "SECRET")
+	req := &ListAPIs{}
+	desc := cs.APIDescription(req)
+	if desc != "lists all available apis on the server" {
+		t.Errorf("APIDescription of listApis is wrong, got %q", desc)
+	}
+}
```
chrisglass: [nitpick]: "\n" missing there, I think.

greut: @chrisglass the lib does it for you there: https://golang.org/src/log/log.go#L170

PR #321, opened 2018-09-13
request: Better error messages in parseResponse

mcorbin:
The Exoscale API can return error messages in the `message` key. Also,
i added the raw message in the error message generated if the response
body is invalid (it will be easier for the users to fire out what
happened exactly).

### DIFF
--- a/request.go
+++ b/request.go
@@ -67,8 +67,13 @@ func (client *Client) parseResponse(resp *http.Response, apiName string) (json.R
 			response, ok = m[key]
 
 			if !ok {
-				for k := range m {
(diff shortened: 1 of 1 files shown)


falzm:
Not sure it is related to your change, but I don't get why the `ok` is evaluated 2 more times (lines 60 and 69) after having evaluated it is false at line 55 🤔 The code would be simpler by removing those 2 extraneous `if`s, IMO.

mcorbin:
This `ok` variable is redefined in every `if` (at each step, we test if the new key exists). But yes, it's possible possible to do it differently.

falzm:
My bad: reviewing code too early is dangerous...

PR #350, opened 2018-10-12
exo: remove it

greut:
- new releases will only be tarballs
- the website won't change (for the time being)
- [x] exoscale/cli v1.0.0 release


pierre-emmanuelJ:
bye bye exo

greut:
(rebased)

PR #353, opened 2018-10-18
travis: run golangci-lint

greut:
less jenkins is good for the skin


greut:
I'll have to write better tests... :crying_cat_face:

PR #355, opened 2018-10-24
async job: don't pick any incoming keys

greut:
The idea is to be more picky when it comes to the jobresult from the async responses because we had some surprises recently.

:warning: **this will break**

### DIFF
--- a/async_jobs.go
+++ b/async_jobs.go
@@ -3,6 +3,9 @@ package egoscale
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
+	"reflect"
+	"strings"
 )
 
 // AsyncJobResult represents an asynchronous job result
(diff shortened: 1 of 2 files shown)


async_jobs.go
```diff
@@ -83,10 +86,18 @@ func (a AsyncJobResult) Result(i interface{}) error {
...
 				}
 
-				// otherwise, pick the first key
-				for k := range m {
-					return json.Unmarshal(m[k], i)
+				typ := reflect.TypeOf(i)
+				key := strings.ToLower(typ.Elem().Name())
+
+				// XXX our setup turned destroy into expunge
+				if a.Cmd == "org.apache.cloudstack.api.command.user.vm.DestroyVMCmd" {
```
pierre-emmanuelJ: 👍

falzm:
Stale, no longer relevant with regards to upcoming API evolution.

PR #357, opened 2018-10-29
cs: move elsewhere

greut:
- move to go modules, no vendoring still as it's a lib
- move `cs` elsewhere. I propose a gitlab/greut repository
- ...
- profit?


greut:
thanks @pyr

PR #359, opened 2018-11-06
global: remove domain and associated fields

greut:
Ignoring fields: `domain`,  `domainid`, `isrecursive`, `iscallerchilddomain`


generate/main.go
```diff
@@ -23,17 +23,19 @@ import (
...
+	"displaynetwork",
+	"displayvm",
+	"displayvolume",
 	"forvpc",
+	"haenable",
+	"hostid",
+	"hypervisor",
+	"iscallerchilddomain",
+	"isdynamicallyscalable",
+	"podid",
```
pierre-emmanuelJ: these field `domainid,domain,regionid,listall,isrecursive...` have not to be here? I don't see it

pierre-emmanuelJ: My bad they are not anymore in  the listApi

greut:
it keeps getting better ;)

```console
> go run generate/main.go -apis listApis.json | awk -F' ' '{ sum += $1 }; END { print sum }'
17
```

PR #363, opened 2018-11-12
Paginate on resource

greut:
- [x] change `Get` semantics

## Making pagination easier

**Before**

```go
zone := &Zone{}
req, _ := zone.ListRequest()
client.Paginate(req, func(i interface{}, e error) bool {
    z := i.(*Zone)
    fmt.Println(zone)
})
```

**After**

```go
client.Paginate(&Zone{}, func(i interface{}, e error) bool {
    z := i.(*Zone)
    fmt.Println(zone)
})
```

## Making List commands more predictable

- `client.Get` will always fail if zero or more than one elements are found using 431/9999.
- `client.List` will never fail if a 431/9999 is returned.

**Before**

```go
zones, err := client.List(Zone{ID: ...})
if e, ok := err.(ErrorResponse); e.ErrorCode == 431 {
    zones = []interface{}{}
}
```

**After**

```go
zones, err := client.List(Zone{ID: ...})
```

**Rationale** it helps dealing with such cases in the Terraform provider [handleNotFound](https://github.com/exoscale/terraform-provider-exoscale/blob/master/exoscale/provider.go#L271-L282)

**NB** CloudStack behaves differently as it doesn't do have _get_, only _list_.


greut:
When on unit test does it all.

```console
% go test -covermode atomic -run TestClientList    
PASS
coverage: 32.2% of statements
```

PR #365, opened 2018-11-15
Add Runstat.us

pierre-emmanuelJ:
- [x] Runstatus.go
- [x] Units test
- [x] Create is said to return 200 and 201 according to the doc... but this might not be true.


runstatus.go
```diff
@@ -0,0 +1,378 @@
...
+// RunstatusServiceList service list
+type RunstatusServiceList struct {
+	Services []RunstatusService `json:"services"`
+}
+
+// ID give the service ID
+func (service *RunstatusService) ID() (int, error) {
+	url := strings.TrimRight(service.URL, "/")
+	urlSplited := strings.Split(url, "/")
+	id, err := strconv.ParseInt(urlSplited[len(urlSplited)-1], 10, 64)
```
greut:
```suggestion
	id, err := strconv.ParseInt(urlSplited[len(urlSplited)-1], 10, 32)
```

runstatus.go
```diff
@@ -0,0 +1,378 @@
...
+
+	var p *RunstatusPageList
+	if err := json.Unmarshal(resp, &p); err != nil {
+		return nil, err
+	}
+
+	return p.Results, nil
+}
+
+func (client *Client) runstatusRequest(ctx context.Context, uri string, params, method string) (json.RawMessage, error) {
```
greut: `params` could be an `interface{}` so `json.Marshal` is only done once.

runstatus.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// DeleteRunstatusMaintenance delete runstatus Maintenance
+func (client *Client) DeleteRunstatusMaintenance(ctx context.Context, page string, id int) error {
+	_, err := client.runstatusRequest(ctx, fmt.Sprintf("/pages/%s/maintenances/%d", page, id), nil, "DELETE")
+	return err
+}
+
+// ID give the maintenance ID
+func (service *RunstatusMaintenance) ID() (int, error) {
```
falzm: I'd name the receiver variable `m` or `maintenance` instead of `service`, since its type is `RunstatusMaintenance`.

runstatus.go
```diff
@@ -0,0 +1,357 @@
...
+	id, err := strconv.ParseInt(urlSplited[len(urlSplited)-1], 10, 64)
+	if err != nil {
+		return 0, err
+	}
+
+	return int(id), nil
+}
+
+// UpdateRunstatusMaintenance update runstatus Maintenance
+func (client *Client) UpdateRunstatusMaintenance(ctx context.Context, page string, id int, event RunstatusEvent) error {
```
falzm: I wonder if it wouldn't be nicer as a method on the `RunstatusMaintenance` type:

```go
func (m *RunstatusMaintenance) Update(ctx context.Context, event RunstatusEvent) error
```

The `page` parameter could be a `RunstatusMaintenance` struct field?

pierre-emmanuelJ: yes it would be nicer But where I put the egoscale `client`, the `page` name to know which page to modify and where I can put the `id` of the maintenance to know which event I have to modify, knowing that id is not naturally in maintenance struct.

runstatus.go
pierre-emmanuelJ: 👍

runstatus.go
pierre-emmanuelJ: `RunstatusMaintenance` cannot let me do that. Or maybe a PR in runstat.us can fix it

greut: I kind of agree with @falzm here. We probably have a HATE-OAS problem in runstatus.

pierre-emmanuelJ: I'm agree too but as you said the problem come from runstatus

falzm: If the Runstat.us data model allows it, I suggest a more fluid API style such as:

```go
type RunstatusPage struct {
        // all current fields, plus:
        client *Client
}

func (client *Client) GetRunstatusPage(ctx context.Context, pageName string) (*RunstatusPage, error) {}

func (page *RunstatusPage) CreateIncident(ctx context.Context, incident RunstatusIncident) error {}

func (page *RunstatusPage) CreateMaintenance(ctx context.Context, maintenance RunstatusMaintenance) error {}

func (m *RunstatusMaintenance) Update(ctx context.Context, event RunstatusEvent) error {}
```

WDTY?

runstatus.go
```diff
@@ -0,0 +1,357 @@
...
+	return p, nil
+}
+
+// DeleteRunstatusPage delete runstatus page
+func (client *Client) DeleteRunstatusPage(ctx context.Context, pageName string) error {
+	_, err := client.runstatusRequest(ctx, "/pages/"+pageName, nil, "DELETE")
+	return err
+}
+
+// GetRunstatusPage delete runstatus page
```
falzm: Function purpose in comment doesn't match function name below

runstatus.go
```diff
@@ -0,0 +1,357 @@
...
+
+	var p *RunstatusPage
+	if err := json.Unmarshal(resp, &p); err != nil {
+		return nil, err
+	}
+
+	return p, nil
+}
+
+// ListRunstatusPage delete runstatus page
```
falzm: Function purpose in comment doesn't match function name below

runstatus.go
pierre-emmanuelJ: 👍 LGTM

runstatus.go
```diff
@@ -198,6 +207,8 @@ func (m *RunstatusMaintenance) ID() (int, error) {
 
 // UpdateRunstatusMaintenance update runstatus Maintenance
 func (client *Client) UpdateRunstatusMaintenance(ctx context.Context, page string, id int, event RunstatusEvent) error {
+	// XXX: GET /pages/$page | jq .maintenances_url
+	//      GEt maintenances_url + "/$id" | jq .events_url
```
greut: @pierre-emmanuelJ some ideas. :1st_place_medal:

pierre-emmanuelJ: LGTM 💯 better to have good URL in reply instead of building them manually. It's safer and more HATE-OAS compliant 👍

pierre-emmanuelJ:
👍 COOL

runstatus.go
```diff
@@ -111,6 +114,8 @@ type RunstatusMaintenance struct {
 	EndDate     *time.Time       `json:"end_date"`
 	Events      []RunstatusEvent `json:"events,omitempty"`
 	EventsURL   string           `json:"events_url,omitempty"`
+	ID          int              `json:"id,omitempty"`       // missing field
```
pierre-emmanuelJ: 👍

runstatus.go
falzm: So, not going in that direction after all?

greut: Embedding the client/page has been done in a more superficial way by hardcoding the `/pages` URL and otherwise using `PageURL`.

```go
client.CreateIncident(ctx, RunstatusIncident{PageURL: page.URL, ...})
// vs
page.CreateIncident(ctx, RunstatusIncident{...})
```

It feels more consistent with the rest of the library.

PR #367, opened 2018-11-19
client: fix authorizeSecurityGroupEgress when non-ptr

greut:
this is suboptimal and we should check that all projects are 1.8+ (I'm looking at you docker/machine)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+0.12.5
+------
+
+- fix: `AuthorizeGroupEgress` could return `authorizeGroupIngress` as name
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+0.12.5
+------
+
+- fix: `AuthorizeGroupEgress` could return `authorizeGroupIngress` as name
```
mcorbin: I don't unterstand why the changelog mentions `authorizeGroupIngress`.

greut: `Authorize..` is the go struct, `authorize...` is the CloudStack API call.

PR #372, opened 2018-12-03
travis: use cc-test-reporter

greut:
it feels like the code coverage is fading away, #365 

**TODO** https://docs.codeclimate.com/docs/travis-ci-test-coverage

```
Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```


greut:
You're the reporter id holder

pyr:
well that clears thigns up

pyr:
Ah, that wasn't obvious at all. i can't find it in the UI though.

greut:
and it's back.

PR #379, opened 2018-12-11
uuid: use gofrs/uuid fork

greut:
Waiting for the next go.uuid release.

https://gof.rs/projects/uuid/

**NB** as we don't vendorize anything in this project, this has to be done again in all the client project, if needed.


greut:
This is useless here.

PR #382, opened 2019-01-16
Runstatus maintenance list

greut:
cf https://github.com/exoscale/cli/pull/84

- `/pages` has an empty maintenances bag, while `/pages/<page>` fills it
- known bug. `/pages/<page>/maintenances` fetches the 50 oldest maintenances
- ditto for incidents and services.


pierre-emmanuelJ:
rebased

PR #388, opened 2019-02-14
affinity group: empty name should be accepted

greut:
```console
% cs createAffinityGroup name="" type="host anti-affinity"
{
  "affinitygroup": {
    "id": "6d06b1e8-f899-4fa9-97ba-1ff202ecc2a4",
    "name": "",
    "type": "host anti-affinity"
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+0.13.4 (unreleased)
+-------------------
+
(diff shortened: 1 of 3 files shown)


mcorbin:
The name seems required, cf https://cloudstack.apache.org/api/apidocs-4.11/apis/createAffinityGroup.html

greut:
The name is required but the value might be empty.

PR #397, opened 2019-04-03
Replace satori/go.uuid by gofrs/uuid

mcorbin:
The satori/go.uuid package has critical
bugs (https://github.com/satori/go.uuid/issues/73).

This commit replaces this package by the gofrs/uuid fork.

In that package, The `Equal` function for UUID has been deleted (we
can use `==` directly for comparisons as explained in the `3.0.0`
release note of the package:
https://github.com/gofrs/uuid/releases/tag/v3.0.0).

Related to https://github.com/containous/traefik/pull/4718

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	uuid "github.com/satori/go.uuid"
+	uuid "github.com/gofrs/uuid"
 )
 
 // UUID holds a UUID v4
@@ -38,7 +38,7 @@ func (u *UUID) DeepCopyInto(out *UUID) {
 
 // Equal returns true if itself is equal to other.
 func (u UUID) Equal(other UUID) bool {
-	return uuid.Equal(u.UUID, other.UUID)
+	return u == other
 }
 
 // UnmarshalJSON unmarshals the raw JSON into the UUID.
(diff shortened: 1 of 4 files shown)


pierre-emmanuelJ:
That's a good point!

uuid.go
```diff
@@ -36,11 +36,6 @@ func (u *UUID) DeepCopyInto(out *UUID) {
 	out.UUID = (uuid.UUID)(o)
 }
 
-// Equal returns true if itself is equal to other.
-func (u UUID) Equal(other UUID) bool {
-	return uuid.Equal(u.UUID, other.UUID)
-}
-
```
greut: I don't think, it's that simple.

mcorbin: why ?

uuid.go
```diff
@@ -38,7 +38,7 @@ func (u *UUID) DeepCopyInto(out *UUID) {
 
 // Equal returns true if itself is equal to other.
 func (u UUID) Equal(other UUID) bool {
-	return uuid.Equal(u.UUID, other.UUID)
+	return u == other
```
pierre-emmanuelJ: Is it working have you tested that?

mcorbin: It should work, cf https://github.com/gofrs/uuid/releases/tag/v3.0.0

pierre-emmanuelJ:
UUID are generated by `Cloudstack` so I think is not relevant to change that now

uuid.go
greut: The embed struct seems to be treated like the type alias. :+1:

mcorbin:
@pierre-emmanuelJ even if UUID are generated by cloudstack, I think it will be nice to get rid of the `satori/go.uuid` package.

PR #400, opened 2019-04-16
Add the RegisterCustomTemplate call

mcorbin:


templates.go
```diff
@@ -13,15 +13,18 @@ type Template struct {
 	Format                string            `json:"format,omitempty" doc:"the format of the template."`
 	HostID                *UUID             `json:"hostid,omitempty" doc:"the ID of the secondary storage host for the template"`
 	HostName              string            `json:"hostname,omitempty" doc:"the name of the secondary storage host for the template"`
+	Hypervisor            string            `json:"hypervisor,omitempty" doc:"the target hypervisor for the template"`
```
greut: wasn't this root only?

templates_test.go
```diff
@@ -33,3 +33,9 @@ func TestDeleteTemplate(t *testing.T) {
 	_ = req.Response().(*AsyncJobResult)
 	_ = req.AsyncResponse().(*BooleanResponse)
 }
+
+func TestRegisterCustomTemplate(t *testing.T) {
+	req := &RegisterCustomTemplate{}
+	_ = req.Response().(*AsyncJobResult)
+	_ = req.AsyncResponse().(*[]Template)
```
greut: pointer to slices are usually overkill.

templates.go
mcorbin: Nope, this field is also visible for non-root users.

pierre-emmanuelJ:
👍

PR #385, opened 2019-01-21
uuid: implements DeepCopy on UUID

greut:
Fixes #384

PR #403, opened 2019-05-12
Add deepCopy on AsyncJobResult

pierre-emmanuelJ:
Add this two methodes for using them in the cluster-api exoscale

cf: https://github.com/exoscale/egoscale/pull/385

### DIFF
--- a/async_jobs.go
+++ b/async_jobs.go
@@ -45,13 +45,9 @@ func (a *AsyncJobResult) DeepCopy() *AsyncJobResult {
 
 // DeepCopyInto copies the receiver into out.
 //
-// In must be non nil.
+// In (a) must be non nil. out must be non nil
 func (a *AsyncJobResult) DeepCopyInto(out *AsyncJobResult) {
-	if a == nil {
-		out = &AsyncJobResult{}
-		return
-	}
-	out = &AsyncJobResult{
+	*out = AsyncJobResult{
 		AccountID:       a.AccountID,
 		Cmd:             a.Cmd,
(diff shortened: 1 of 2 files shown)


async_jobs.go
```diff
@@ -21,6 +21,52 @@ type AsyncJobResult struct {
...
+
+	return &AsyncJobResult{
+		AccountID:       a.AccountID,
+		Cmd:             a.Cmd,
+		Created:         a.Created,
+		JobID:           a.JobID,
+		JobInstanceID:   a.JobInstanceID,
+		JobInstanceType: a.JobInstanceType,
+		JobProcStatus:   a.JobProcStatus,
+		JobResult:       a.JobResult,
```
greut:
```suggestion
		JobResult:       a.JobResult.DeepCopy(),
```

async_jobs.go
```diff
@@ -21,6 +21,48 @@ type AsyncJobResult struct {
...
+		JobStatus:       a.JobStatus,
+		UserID:          a.UserID.DeepCopy(),
+	}
+}
+
+// DeepCopyInto copies the receiver into out.
+//
+// In (a) must be non nil. out must be non nil
+func (a *AsyncJobResult) DeepCopyInto(out *AsyncJobResult) {
+	*out = AsyncJobResult{
```
mcorbin: Why not use `a.DeepCopy() here ?

pierre-emmanuelJ: `// In (a) must be non nil. out must be non nil`
we want that (program crash), not nil result

PR #404, opened 2019-05-31
Improve the requests User-Agent HTTP header

falzm:
This change makes the "User-Agent" HTTP request header more informative,
and exposes it outside of the package if a caller wants to prefix it
with is own UA segment:

```go
import "github.com/exoscale/egoscale"

...

egoscale.UserAgent = fmt.Sprintf("Exoscale-CLI/%s %s", version, egoscale.UserAgent)

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+UNRELEASED
+----------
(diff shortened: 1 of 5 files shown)


falzm:
[ch3708]

PR #411, opened 2019-10-08
Add IAM

illi-j:
This PR add IAM to egoscale.

### DIFF
--- a/iam_apikey.go
+++ b/iam_apikey.go
@@ -15,7 +15,7 @@ type APIKey struct {
 	Description string     `json:"description"`
 	Key         string     `json:"key"`
 	Secret      string     `json:"secret,omitempty"`
-	Operations  string     `json:"operations,omitempty"`
+	Operations  []string   `json:"operations,omitempty"`
 	Type        APIKeyType `json:"type"`
 }
 
@@ -38,8 +38,8 @@ type ListAPIKeys struct {
 
 // ListAPIKeysResponse represents a list of apikeys
 type ListAPIKeysResponse struct {
-	Count      int      `json:"count"`
-	ListAPIKey []APIKey `json:"apikey"`
+	Count   int      `json:"count"`
+	APIKeys []APIKey `json:"apikeys"`
 }
 
 // Response returns the struct to unmarshal
--- a/iam_apikey.go
+++ b/iam_apikey.go
@@ -1,15 +1,5 @@
 package egoscale
 
-import (
-	"context"
-	"encoding/json"
-	"fmt"
-	"io/ioutil"
-	"net/http"
-	"net/url"
-	"strings"
-)
-
 // APIKeyType holds the type of the apikey
 type APIKeyType string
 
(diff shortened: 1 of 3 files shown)


iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+	Key         string     `json:"key"`
+	Secret      string     `json:"secret,omitempty"`
+	Operations  []string   `json:"operations,omitempty"`
+	Type        APIKeyType `json:"type"`
+}
+
+// CreateAPIKey represents the apikey creation
+type CreateAPIKey struct {
+	Description string `json:"description"`
+	Operations  string `json:"operations,omitempty"`
```
pierre-emmanuelJ: I think can be nice to put a slice for this field

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
+package egoscale
+
+// APIKeyType holds the type of the apikey
+type APIKeyType string
+
+const (
+	// APIKeyUnrestricted is unrestricted
+	APIKeyUnrestricted APIKeyType = "restricted"
+	// APIKeyRestricted is restricted
+	APIKeyRestricted APIKeyType = "unrestricted"
```
falzm: I suggest naming the constants `APIKeyTypeRestricted`/`APIKeyTypeUnrestricted` so one can know at a glance what they refer to.

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// APIKey represents an apikey
```
falzm:
```suggestion
// APIKey represents an API key
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
+package egoscale
+
+// APIKeyType holds the type of the apikey
```
falzm:
```suggestion
// APIKeyType holds the type of the API key
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+// APIKey represents an apikey
+type APIKey struct {
+	Description string     `json:"description"`
+	Key         string     `json:"key"`
+	Secret      string     `json:"secret,omitempty"`
+	Operations  []string   `json:"operations,omitempty"`
+	Type        APIKeyType `json:"type"`
+}
+
+// CreateAPIKey represents the apikey creation
```
falzm:
```suggestion
// CreateAPIKey represents the API key creation
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+	Operations  string `json:"operations,omitempty"`
+	_           bool   `name:"createApiKey" description:"Create an apikey."`
+}
+
+// Response returns the struct to unmarshal
+func (CreateAPIKey) Response() interface{} {
+	return new(APIKey)
+}
+
+// ListAPIKeys represents a search for an apikeys
```
falzm:
```suggestion
// ListAPIKeys represents a search for an API key
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+type ListAPIKeys struct {
+	_ bool `name:"listApiKeys" description:"List apikeys."`
+}
+
+// ListAPIKeysResponse represents a list of apikeys
```
falzm:
```suggestion
// ListAPIKeysResponse represents a list of API keys
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+	Count   int      `json:"count"`
+	APIKeys []APIKey `json:"apikeys"`
+}
+
+// Response returns the struct to unmarshal
+func (ListAPIKeys) Response() interface{} {
+	return new(ListAPIKeysResponse)
+}
+
+// RevokeAPIKey represents a revoke of apikey
```
falzm:
```suggestion
// RevokeAPIKey represents a revocation of an API key
```

iam_apikey.go
```diff
@@ -0,0 +1,64 @@
...
+type RevokeAPIKey struct {
+	Key string `json:"key"`
+	_   bool   `name:"revokeApiKey" description:"Revoke an apikey."`
+}
+
+// RevokeAPIKeyResponse represents a revoke of apikey
```
falzm:
```suggestion
// RevokeAPIKeyResponse represents the response to an API key revocation
```

iam_apikey.go
falzm: I agree.

falzm:
If you PR is ready to be reviewed you can remove the WIP label ;)

iam_apikey.go
greut: Also, they are obviously swapped.

iam_apikey.go
illi-j: not possible for the API at the moment.
(likely to change in the near future)

iam_apikey.go
```diff
@@ -47,10 +47,25 @@ func (ListAPIKeys) Response() interface{} {
 	return new(ListAPIKeysResponse)
 }
 
+// ListAPIKeyOperations represents a search for operations for the current API key
```
falzm: Is it really necessary to implement this call? Doesn't the `getAPIKey` also return the key's supported operations?

alexanderjamesking: There is a difference.

ListApiKeyOperations returns all of the things a key can do (even for a key without restrictions - you get the full list), whereas GetApiKey returns the details of the policy for the key - so for an unrestricted key no operations would be returned.

ListApiKeyOperations is primarily for the UI so we can display a list of checkboxes when creating a new key, it's debatable how useful this is from the CLI as the operations do not always align with the commands exposed by the CLI. If we were to do some autocomplete on the operations from the CLI you could use it.

PR #424, opened 2020-03-11
Egoscale: add SetTLSAuthentication in client.go to support tls

illi-j:


client.go
```diff
@@ -103,6 +105,36 @@ func NewClient(endpoint, apiKey, apiSecret string) *Client {
...
+	caCertPool.AppendCertsFromPEM(caCert)
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{cert},
+		RootCAs:      caCertPool,
+	}
+	tlsConfig.BuildNameToCertificate()
+
+	httpClient := &http.Client{
+		Transport: &http.Transport{TLSClientConfig: tlsConfig},
```
falzm: By proceeding this way you overwrite a potentially configured Transport – e.g. if the user already has called [`Client.TraceOn()`](https://godoc.org/github.com/exoscale/egoscale#Client.TraceOn). You should instead call [`http.Transport.Clone()`](https://godoc.org/net/http#Transport.Clone) and then set the `TLSClientConfig` on the cloned transport, something like:

```go
tlsTransport := client.HTTPClient.Transport.Clone()
tlsTransport.TLSClientConfig = tlsConfig
client.HTTPClient = tlsTransport
```

(haven't tested it though 😉)

greut:
Since `HTTPClient` is exposed, there is probably no needs to bloat this library with that. E.g. the usage in the Terraform provider, https://github.com/terraform-providers/terraform-provider-exoscale/blob/master/exoscale/client.go#L35

PR #425, opened 2020-03-23
Add support for public API V2

falzm:
This change introduces public API V2 support, for which low-level calls are code-generated from our OpenAPI Specification (OAS) in an internal package and then wrapped in higher level calls at the root of the egoscale package (see the new `Client.Ping()` method for implementation example).


client.go
```diff
@@ -100,6 +107,22 @@ func NewClient(endpoint, apiKey, apiSecret string) *Client {
 		client.TraceOn()
 	}
 
+	exoSecurityProvider, err := v2.NewSecurityProviderExoscaleV2(client.APIKey, client.apiSecret)
+	if err != nil {
+		panic(errors.Wrap(err, "unable to initialize security provider"))
```
jeromer: maybe panicking is a bit brutal here. Is there a way to return an explicit error without panicking ?

client.go
```diff
@@ -100,6 +107,22 @@ func NewClient(endpoint, apiKey, apiSecret string) *Client {
...
+	exoSecurityProvider.ReqExpire = client.Expiration
+
+	opts := []v2.ClientOption{
+		v2.WithHTTPClient(client.HTTPClient),
+		v2.WithBaseURL(client.Endpoint),
+		v2.WithRequestEditorFn(exoSecurityProvider.Intercept),
+	}
+
+	if client.v2, err = v2.NewClientWithResponses(client.Endpoint, opts...); err != nil {
+		panic(errors.Wrap(err, "unable to initialize API client"))
```
jeromer: same here

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// NewSecurityProviderExoscaleV2 returns a new Exoscale API V2 security provider to sign API requests using the
+// specified API key/secret.
+func NewSecurityProviderExoscaleV2(apiKey, apiSecret string) (*SecurityProviderExoscaleV2, error) {
+	var provider = SecurityProviderExoscaleV2{
+		ReqExpire: 10 * time.Minute,
+	}
+
+	if provider.apiKey = apiKey; provider.apiKey == "" {
```
jeromer: This syntaxic sugar is hard to process. Is there a way to make it more explicit ?

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, errors.New("missing API key")
+	}
+
+	if provider.apiSecret = apiSecret; provider.apiSecret == "" {
```
jeromer: same here

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	if err := buf.WriteByte('\n'); err != nil {
+		return err
+	}
+
+	// Request query string parameters
+	// Important: this is order-sensitive, we have to have to sort parameters alphabetically to ensure signed
+	// values match the names listed in the "signed-query-args=" signature pragma.
+	signedParams := make([]string, 0)
```
jeromer:
```suggestion
	signedParams := []string{}
```

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+	var buf bytes.Buffer
+
+	authHeader := fmt.Sprintf("EXO2-HMAC-SHA256 credential=%s", s.apiKey)
+
+	// Request method/URL path
+	if _, err := buf.WriteString(fmt.Sprintf("%s %s\n", req.Method, req.URL.Path)); err != nil {
+		return err
+	}
+
+	// Request body
```
jeromer: I'd move this into a dedicated function. `slurpRequestBody` maybe ?

client.go
falzm: Sadly no, because it would require us to change the `NewClient()` function signature to return an `error` which would break all implementations (we have to assume egoscale is used outside of our official tooling). If you think of a better way, feel free to suggest :)

client.go
falzm: Same answer.

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		if _, err := buf.Write(body); err != nil {
+			return err
+		}
+	}
+	if err := buf.WriteByte('\n'); err != nil {
+		return err
+	}
+
+	// Request query string parameters
```
jeromer: Same here, maybe a dedicated function like `extractSignedParams` would be more suitable

internal/v2/security.go
falzm: You mean something like

```go
provider.apiKey = apiKey
if provider.apiKey == "" {
	// ...
```

?

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+// Intercept is an HTTP middleware that intercepts and signs client requests before sending them to the API
+// endpoint.
+func (s *SecurityProviderExoscaleV2) Intercept(_ context.Context, req *http.Request) error {
+	return s.signRequest(req, time.Now().UTC().Add(s.ReqExpire))
+}
+
+func (s *SecurityProviderExoscaleV2) signRequest(req *http.Request, expiration time.Time) error {
+	var buf bytes.Buffer
+
+	authHeader := fmt.Sprintf("EXO2-HMAC-SHA256 credential=%s", s.apiKey)
```
jeromer: Instead of concatenating strings in `authHeader` how about doing something like `authHeaderParts = []string{}` then you append into this array whenever needed. And once you are done with it you simple run `strings.Join(authHeaderParts, ',')` which will avoid any mistake regarding the `,` separator ?

internal/v2/security.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	authHeader += fmt.Sprintf(",expires=%d", expiration.Unix())
+
+	h := hmac.New(sha256.New, []byte(s.apiSecret))
+	if _, err := h.Write(buf.Bytes()); err != nil {
+		return err
+	}
+	authHeader += ",signature=" + base64.StdEncoding.EncodeToString(h.Sum(nil))
+
+	req.Header.Set("Authorization", authHeader)
```
jeromer: And then here, something like:
```
req.Header.Set("Authorization", strings.Join(authHeaderParts, ","))
```

internal/v2/security.go
falzm: That's has also been suggested in https://github.com/exoscale/requests-exoscale-auth/pull/2 so you're right: I should use the same method here.

client.go
jeromer: > Sadly no, because it would require us to change the NewClient() function signature to return an error which would break all implementations (we have to assume egoscale is used outside of our official tooling).

Indeed.

client.go
jeromer: same indeed :)

internal/v2/security.go
jeromer: Something like:
```go
if apiKey == "" {
return nil, errors.New("missing API key")
}

if apiSecret == "" {
return nil, errors.New("missing API key")
}

provided.apiKey = apiKey
provider.apiSecret = apiSecret

return &provider, nil
```

falzm: Changed in 5aa5488.

internal/v2/security.go
falzm: Also changed in 5aa5488.

internal/v2/security.go
falzm: Changed in f0d8769.

internal/v2/security.go
falzm: Changed in f0d8769.

internal/v2/security.go
falzm: The change to concatenating signature parts instead of manually writing to a buffer reduced the global volume of code to a point where I find it overkill to extract this to a dedicated function, especially since it's not reused anywhere...

internal/v2/security.go
falzm: Changed in f0d8769.

internal/v2/security.go
jeromer: You have applied the part thing method for `sigParts` but not for `authHeader` whereas it seems it gets also concatenated at some point. It there a reason for this ?

falzm: Nope that was an oversight and you're right, I'll get to it right away 😳

falzm: Fixed in 289d51a.

PR #426, opened 2020-04-16
signature: copy body when read

pyr:
http.Request offers GetBody() to provide a body copy,
but it is not available by the time we get to the signature
process in some contexts

### DIFF
--- a/internal/v2/security.go
+++ b/internal/v2/security.go
@@ -1,6 +1,7 @@
 package v2
 
 import (
+	"bytes"
 	"context"
 	"crypto/hmac"
 	"crypto/sha256"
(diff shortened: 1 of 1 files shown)


internal/v2/security.go
```diff
@@ -57,16 +57,15 @@ func (s *SecurityProviderExoscaleV2) signRequest(req *http.Request, expiration t
...
-	body := ""
+	// Request body if present
 	if req.Body != nil {
 		data, err := ioutil.ReadAll(req.Body)
 		if err != nil {
 			return err
 		}
-		body = string(data)
+		sigParts = append(sigParts, string(data))
+		req.Body = newCopiedBody(data)
```
7fELF, line 67:
```suggestion
		req.Body = ioutil.NopCloser(bytes.NewReader(data))
```

pierre-emmanuelJ: see my commit reference in the PR if you want to try

pyr: Thanks :fireworks:

PR #427, opened 2020-04-22
add support for snapshot export

illi-j:


snapshots.go
```diff
@@ -137,3 +137,19 @@ func (RevertSnapshot) Response() interface{} {
 func (RevertSnapshot) AsyncResponse() interface{} {
 	return new(BooleanResponse)
 }
+
+// ExportSnapshot exports an instant snapshot of a volume to SOS
```
falzm, line 141:
```suggestion
// ExportSnapshot exports a volume snapshot.
```

snapshots.go
```diff
@@ -137,3 +137,19 @@ func (RevertSnapshot) Response() interface{} {
...
+type ExportSnapshot struct {
+	ID *UUID `json:"id"`
+	_  bool  `name:"exportSnapshot" description:"Exports an instant snapshot of a volume."`
+}
+
+// ExportSnapshotResponse represents the presigned URL to access the exported snapshot
```
falzm, line 147: There will be more information returned besides the pre-signed URL at some point (e.g. the file checksum).

```suggestion
// ExportSnapshotResponse represents the response of a snapshot export operation.
```

snapshots.go
```diff
@@ -137,3 +137,19 @@ func (RevertSnapshot) Response() interface{} {
 func (RevertSnapshot) AsyncResponse() interface{} {
 	return new(BooleanResponse)
 }
+
+// ExportSnapshot exports a volume snapshot
+type ExportSnapshot struct {
+	ID *UUID `json:"id"`
+	_  bool  `name:"exportSnapshot" description:"Exports an instant snapshot of a volume."`
```
pierre-emmanuelJ, line 144:
```suggestion
	_  bool  `name:"exportSnapshot" description:"Exports an instance snapshot of a volume."`
```
instance?

falzm: @pierre-emmanuelJ "instant" (referring to the snapshot of a instantaneous capture of a state) is correct, although a bit overkill but it's used elsewhere in other calls in this package so ¯\\\_(ツ)_/¯

pierre-emmanuelJ: Okay I got it ;)

PR #432, opened 2020-05-28
Add support for Network Load Balancer resources

falzm:
This change introduces support for Network Load Balancer (NLB) resources
management.


falzm:
[ch13590]

internal/v2/loadbalancer.go
```diff
@@ -0,0 +1,40 @@
+package v2
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// UnmarshalJSON unmarshals a LoadBalancer resource into a temporary structure whose "CreatedAt" field of type
+// string to be able to parse the original timestamp (ISO 8601) into a time.Time object, since json.Unmarshal()
+// only supports RFC3339 format.
```
greut, line 10: Alternatively, you could have created your own `ISO8601Time`, https://github.com/exoscale/egoscale/blob/master/uuid.go#L44-L61

illi-j:
IHMO, I think using `uint` for everything related to ports would be nice

falzm:
> IHMO, I think using `uint` for everything related to ports would be nice

My bad, I remember we've talked about this indeed. That'll be `uint16`, actually ;)

falzm:
> > IHMO, I think using `uint` for everything related to ports would be nice
> 
> My bad, I remember we've talked about this indeed. That'll be `uint16`, actually ;)

Fixed, you may update your vendored modules to retrieve the latest version @illi-j @pierre-emmanuelJ.

PR #435, opened 2020-06-24
Fix a bug in the ListNetworkLoadBalancers call

falzm:
This changes fixes a bug in the `ListNetworkLoadBalancers()` call that
didn't set the embedded client/zone fields and lead to a nil pointer
derefencing when calling methods on a `NetworkLoadBalancerService`
object.

### DIFF
--- a/loadbalancer.go
+++ b/loadbalancer.go
@@ -312,10 +312,8 @@ func (c *Client) ListNetworkLoadBalancers(ctx context.Context, zone string) ([]*
 	}
 
 	if resp.JSON200.LoadBalancers != nil {
-		for _, n := range *resp.JSON200.LoadBalancers {
-			n := n
-
-			nlb := nlbFromAPI(&n)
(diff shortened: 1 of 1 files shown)


loadbalancer.go
```diff
@@ -317,9 +312,14 @@ func (c *Client) ListNetworkLoadBalancers(ctx context.Context, zone string) ([]*
 	}
 
 	if resp.JSON200.LoadBalancers != nil {
-		for _, nlb := range *resp.JSON200.LoadBalancers {
-			nlb := nlb
-			list = append(list, nlbFromAPI(&nlb))
+		for _, n := range *resp.JSON200.LoadBalancers {
+			n := n
+
+			nlb := nlbFromAPI(&n)
```
pierre-emmanuelJ, line 318:
```suggestion
		for i := range *resp.JSON200.LoadBalancers {
			nlb := nlbFromAPI(&(*resp.JSON200.LoadBalancers)[i])
```

PR #437, opened 2020-07-09
Add evictInstancePoolMembers call

pierre-emmanuelJ:
Add `evictInstancePoolMembers` call

### DIFF
--- a/instance_pool.go
+++ b/instance_pool.go
@@ -1,6 +1,6 @@
 package egoscale
 
-// InstancePoolState represents the state of an instance pool.
+// InstancePoolState represents the state of an Instance Pool.
 type InstancePoolState string
 
 const (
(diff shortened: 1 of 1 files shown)


instance_pool.go
```diff
@@ -83,82 +83,80 @@ type UpdateInstancePool struct {
...
+// Response returns an empty structure to unmarshal an Instance Pool list API response into.
 func (ListInstancePools) Response() interface{} {
 	return new(ListInstancePoolsResponse)
 }
+
+// EvictInstancePoolMembers represents an Instance Pool members eviction API request.
+type EvictInstancePoolMembers struct {
+	ID        *UUID  `json:"id"`
+	ZoneID    *UUID  `json:"zoneid"`
+	MemberIDs []UUID `json:"memberids"`
```
falzm, line 155: Just to be consistent across all fields of the struct:

```suggestion
	MemberIDs []*UUID `json:"memberids"`
```

pierre-emmanuelJ: In this case a slice of pointer is not needed ;) let's stay consistent with the rest of the structure request, like
`UpdateVMAffinityGroup`...etc and many others.

PR #442, opened 2020-08-11
ListVirtualMachines: add new parameter ManagerID

falzm:
This changes introduces a new `ManagerID` filtering parameter to the
`ListVirtualMachines` call, enabling users to list the Compute instances
belonging to a specific Instance Pool.


falzm:
@exo-cedric this is the terminology from the API itself:

```console
$ cs listVirtualMachines managerid=6dbc5d18-981e-7ae4-b6ff-542d81c4d347 fetch_list=True | jq -r '.[].managerid'
6dbc5d18-981e-7ae4-b6ff-542d81c4d347
```

We'll have an opportunity to choose a better one soon with the API V2 😉

PR #450, opened 2020-09-17
Improve GetWithContext() error messages.

falzm:
This change improves the error messages returned by the `GetWithContext`
method in case no resource or multiple resources are found, as it was
previously returned an error suggesting an API-side issue.

### DIFF
--- a/client.go
+++ b/client.go
@@ -150,8 +150,11 @@ func (client *Client) GetWithContext(ctx context.Context, ls Listable) (interfac
 		return nil, err
 	}
 
-	count := len(gs)
-	if count != 1 {
+	switch len(gs) {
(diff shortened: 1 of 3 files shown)


pierre-emmanuelJ:
Good idea 👍 , I'm agreed with this. But as I see, we lose the payload information to tell the caller which resource we are trying to get.

falzm:
> Good idea 👍 I'm agreed with this. But as I see, we lose the payload information to tell the caller which resource we are trying to get.

The caller knows which resource it's trying to get 😏

pierre-emmanuelJ (approved):
LGTM, Thanks!

bcachet (approved):

PR #455, opened 2020-11-26
Improve API v2 errors handling

falzm:
This change introduces a custom `v2.HttpRequestDoer` interface
implementation wrapping the client's `http.Client`, allowing us to
extract relevant error messages from API responses and report them to
the caller.

### DIFF
--- a/client.go
+++ b/client.go
@@ -2,6 +2,7 @@ package egoscale
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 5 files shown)


falzm:
With this change, a client can now have more information about the nature of an error, for example:

```go
cluster, err := c.GetSKSCluster(ctx, zone, "blah")
if err != nil {
	switch {
	case errors.Is(err, egoscale.ErrInvalidRequest):
		panic(fmt.Sprintf("!!!: %s", err))
	case errors.Is(err, egoscale.ErrAPIError):
		panic(fmt.Sprintf("~~~: %s", err))
	default:
		panic(fmt.Errorf("???: %s", err))
	}
}
```

mcorbin (approved):

illi-j (approved):
LGTM!

PR #462, opened 2020-12-16
Fix UpdateIPAddress HealthcheckTLSSkipVerify field

falzm:
This change fixes a bug where the `UpdateIPAddress` operation would
always send `tls-skip-verify: false` in case the
`HealthcheckTLSSkipVerify` field value is not `true`, causing an
existing EIP healthcheck configuration to be modified inadvertently.


mcorbin (approved):

falzm:
[ch23546]

PR #463, opened 2020-12-21
SKS: add new methods

falzm:
This change introduces new SKS-related methods.

### DIFF
--- a/sks.go
+++ b/sks.go
@@ -212,6 +212,29 @@ func (c *SKSCluster) ScaleNodepool(ctx context.Context, np *SKSNodepool, nodes i
 	return nil
 }
 
+// EvictNodepoolMembers evicts the specified members (identified by their Compute instance ID) from the
+// SKS cluster Nodepool.
+func (c *SKSCluster) EvictNodepoolMembers(ctx context.Context, np *SKSNodepool, members []string) error {
+	resp, err := c.c.v2.EvictSksNodepoolMembersWithResponse(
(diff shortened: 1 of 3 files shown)


sgmac:
This looks very good :+1:   I don't think I could say that much honestly, I need to get familiar with `httpmock`. Thanks for adding me to the review.

sgmac (approved):
I tested your branch, all good. The CI passed tests, so there's not that much I can help. Feel free to add me to more reviews. Thanks.

falzm:
Thanks for your feedback @sgmac ;)

sks.go
```diff
@@ -212,6 +212,36 @@ func (c *SKSCluster) ScaleNodepool(ctx context.Context, np *SKSNodepool, nodes i
...
 
+// EvictNodepoolMembers evicts the specified members (identified by their Compute instance ID) from the
+// SKS cluster Nodepool.
+func (c *SKSCluster) EvictNodepoolMembers(ctx context.Context, np *SKSNodepool, members []string) error {
+	instances := make([]string, len(members))
+
+	for i := range members {
+		id := members[i]
+		instances[i] = id
+	}
```
illi-j, line 223: I don't understand the goal of this part
why not use `members` directly?

pierre-emmanuelJ (approved):

sks.go
falzm: You're right, this isn't necessary – fixed in 30b5016.

illi-j (approved):

PR #466, opened 2021-01-19
Improve API v2 async job tests and error reporting

falzm:
This change improves the v2 API async jobs error reporting by returning
the original context error back to the caller so they can use
`errors.Is/As()` and react to a timeout or cancellation more
appropriately.

Also, async poller tests have been split into scenario-dedicated
functions for a better readability.


pierre-emmanuelJ (approved):

internal/v2/async_test.go
```diff
@@ -34,62 +34,50 @@ func TestPoller_WithTimeout(t *testing.T) {
...
-			5*time.Second,
-			time.Second,
-			"polling must NOT complete before the timeout")
-	}
+func TestPoller_Poll_FinishWithError(t *testing.T) {
+	poller := NewPoller().WithInterval(time.Second)
+	require.Eventually(t,
+		func() bool {
+			_, err := poller.Poll(context.Background(),
+				newTestMockPollFunc(time.Second, true, nil, errors.New("o noes")))
```
illi-j, line 58: :D

illi-j (approved):

PR #480, opened 2021-03-10
SKS: add SKSCluster.AuthorityCert() method

falzm:
This change adds a new `AuthorityCert()` method to the `SKSCluster`
struct.


falzm:
[ch27178]

PR #489, opened 2021-05-12
v2: add support for Deploy Target/Instance Prefix to SKS Nodepools

falzm:
* Add `Instance.(At|De)tach*` methods
* Add support for Instance Prefix to SKS Nodepools
* Add support for Deploy Targets to SKS Nodepools


falzm:
[ch30161]

PR #525, opened 2021-09-09
Add support for external-sources

pyr:


exo-cedric (approved):
LGTM

(can't we get rid of the `oapi.NewPoller()` duplication ?)

falzm:
> LGTM
> 
> (can't we get rid of the `oapi.NewPoller()` duplication ?)

What duplication? Those are 2 different execution scopes (2 different methods).

pyr:
> What duplication? Those are 2 different execution scopes (2 different methods).

You mean reducing this to something more succinct since it's code that lives in many of these methods?

```go
	_, err = oapi.NewPoller().
		WithTimeout(c.timeout).
		WithInterval(c.pollInterval).
		Poll(ctx, c.OperationPoller(zone, *resp.JSON200.Id))
	if err != nil {
		return err
	}
```

If so, I don't think it would belong in this PR since it is the way the whole code base is structured, deviating here would introduce an exception for these two calls.

PhilippeChepy (approved):

falzm (approved):
Cosmetic edits, otherwise LGTM 👍

pyr:
Thanks @falzm  for helping with the tests!

PR #539, opened 2021-11-08
v2: add ListInstancesOpt options

falzm:
This change introduces a new `ListInstancesOpt` type to pass options to
the `Client.ListInstances()` method.


falzm:
[sc-37209]

PR #540, opened 2021-11-09
v2: add support for IAM access keys management

falzm:
This change introduces support for IAM access keys management via
the Exoscale V2 API.


falzm:
[sc-37427]

PR #553, opened 2022-03-08
v2: refresh openapi generated code + fix missing type oapi.Reference

7fELF:


v2/oapi/oapi.gen.go
```diff
@@ -2062,18 +2099,6 @@ type Quota struct {
 	Usage *int64 `json:"usage,omitempty"`
 }
 
-// Resource reference
-type Reference struct {
```
exo-cedric, line 2066: Mmmh. I guess this is the oapi.Reference type that is now missing. But then, shouldn't it be defined somewhere else (in the v2 code base) and be used in place of all those struct { ... } ?

7fELF: Even if I recreate the type in the oapi package, it will not accept it, because it is not the type it wants.
```
./anti_affinity_group_test.go:46:6: cannot use &oapi.Reference{...} (type *oapi.Reference) as type *struct { Command *string "json:\"command,omitempty\""; Id *string "json:\"id,omitempty\""; Link *string "json:\"link,omitempty\"" } in field value
```

So I would have to use a type conversion that would look like that:
```go
&oapi.Operation{
	Id: &testOperationID,
	Reference: (*struct {
		Command *string `json:"command,omitempty"`
		Id      *string `json:"id,omitempty"` // revive:disable-line
		Link    *string `json:"link,omitempty"`
	})(&oapi.Reference{Id: &testAntiAffinityGroupID}),
	State: &testOperationState,
}
```

exo-cedric (approved):

dramaturg (approved):

v2/oapi/oapi.gen.go
7fELF: I reduced the boilerplate a bit by doing this in a function:

https://github.com/exoscale/egoscale/blob/682f8278ef5a28f90b72644858f86b51446ea9c4/v2/oapi/reference.go#L3:L18

PR #563, opened 2022-09-12
v2: silence go-retryablehttp client

kobajagi:
This PR prevents `go-retryablehttp` client from writing debug logs to stderr.

### DIFF
--- a/v2/client.go
+++ b/v2/client.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"log"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


v2/client.go
```diff
@@ -31,7 +33,12 @@ var UserAgent = fmt.Sprintf("egoscale/%s (%s; %s/%s)",
 
 // defaultHTTPClient is HTTP client with retry logic.
 // Default retry configuration can be found in go-retryablehttp repo.
-var defaultHTTPClient = retryablehttp.NewClient().StandardClient()
+var defaultHTTPClient = func() *http.Client {
+	rc := retryablehttp.NewClient()
+	// silence client by default
+	rc.Logger = log.New(io.Discard, "", 0)
```
zyegfryed, line 39: Does it mean ALL logs will be discarded? Even the error/panic ones?

7fELF: The methods of the client should return errors

7fELF (approved):

v2/client.go
kobajagi: Generic errors (eg. HTTP error code) will propagate down to egoscale same way as with standard HTTP client, but some `go-retryablehttp` specific error context is lost (eg. failed after nth retry attempt).

What we really need in egoscale is leveled logger to separate debug logs from errors. This is supported by `go-retryablehttp` (check Logger field):
https://pkg.go.dev/github.com/hashicorp/go-retryablehttp?utm_source=godoc#Client

It would be a separate effort to implement this in egoscale.

PhilippeChepy (approved):

PR #570, opened 2022-12-21
template: GetTemplateByName and sorting for []*Template

jessicatoscani:
```
go test -timeout 30s -testify.m ^(TestClient_GetTemplateByName)$ github.com/exoscale/egoscale/v2

PASS
ok  	github.com/exoscale/egoscale/v2	3.015s

```

### DIFF
--- a/v2/template.go
+++ b/v2/template.go
@@ -2,6 +2,7 @@ package v2
 
 import (
 	"context"
+	"sort"
 	"time"
 
 	apiv2 "github.com/exoscale/egoscale/v2/api"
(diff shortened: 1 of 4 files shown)


v2/template.go
```diff
@@ -29,6 +30,22 @@ type Template struct {
 	Zone            *string
 }
 
+// Implement sort.Interface for []*Template by CreatedAt or by Nane
+type Templates []*Template
+
+func (s Templates) Len() int      { return len(s) }
+func (s Templates) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
```
greut, line 37: Just use generics, https://pkg.go.dev/golang.org/x/exp/slices#Sort

arapaho (approved):

v2/template.go
PhilippeChepy: This project is still buildable with go 1.17 for now; generics are available from 1.18.x

PhilippeChepy (approved):

v2/template.go
greut: 1.20 is around the corner. It's likely you won't lose anything by bumping it to 1.18.

jessicatoscani: We will do it later

PR #571, opened 2022-12-28
template: add FindTemplate and fix GetTemplateByName

jessicatoscani:


v2/template.go
```diff
@@ -165,7 +165,21 @@ func (c *Client) GetTemplateByName(ctx context.Context, zone string, templateNam
...
-	return nil, err
+	return nil, apiv2.ErrNotFound
+}
+
+// FindTemplate attempts to find a template by name or ID.
+// In case the identifier is a name and multiple resources match, the newest template is returned.
+func (c *Client) FindTemplate(ctx context.Context, zone, x string, visibilty string) (*Template, error) {
+	// Opportunistic shortcut in case the template is referenced by ID.
+	template, err := c.GetTemplate(ctx, zone, x)
+	if err != nil {
```
kobajagi, line 176: We should split the logic according to the error returned:
- 404 (not found) => look by name
- any other error (eg. 500) => return error

You can check if it is 404 using `errors.Is(err, apiv2.ErrNotFound)`

v2/template.go
```diff
@@ -165,7 +165,21 @@ func (c *Client) GetTemplateByName(ctx context.Context, zone string, templateNam
...
+		template, err = c.GetTemplateByName(ctx, zone, x, visibilty)
+		if err != nil {
+			return nil, apiv2.ErrNotFound
```
kobajagi, line 179: We will lose the correct error message here and it will be a pain to debug later. Just return `err` instead of `apiv2.ErrNotFound`.

v2/template.go
jessicatoscani: https://github.com/exoscale/egoscale/pull/571/commits/f63594a605131d7aedd108d581fdcd81b40643e6 (we receive ErrInvalidRequest when we do not provide a correct uuid to GetTemplate)

kobajagi: Maybe check for both just to be on the safe side?
```
if errors.Is(err, apiv2.ErrInvalidRequest) || errors.Is(err,  apiv2.ErrNotFound) {
```

kobajagi (approved):

PhilippeChepy (approved):
:+1:

PR #578, opened 2023-04-07
sg-rule: support public security groups when listing

pyr:
Rules should have a simple discriminant to understand when a public security group is being referred to

### DIFF
--- a/v2/security_group_rule.go
+++ b/v2/security_group_rule.go
@@ -12,16 +12,18 @@ import (
 
 // SecurityGroupRule represents a Security Group rule.
 type SecurityGroupRule struct {
-	Description     *string
-	EndPort         *uint16
-	FlowDirection   *string `req-for:"create"`
-	ICMPCode        *int64
-	ICMPType        *int64
-	ID              *string `req-for:"delete"`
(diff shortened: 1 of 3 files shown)


PhilippeChepy (approved):

kobajagi (approved):

v2/security_group_rule.go
```diff
@@ -56,18 +58,33 @@ func securityGroupRuleFromAPI(r *oapi.SecurityGroupRule) *SecurityGroupRule {
...
+			if r.SecurityGroup != nil &&
+				(r.SecurityGroup.Visibility == nil ||
+					*r.SecurityGroup.Visibility != "public") {
 				v = r.SecurityGroup.Id
 			}
 			return
 		}(),
+		SecurityGroupName: func() (v *string) {
+			if r.SecurityGroup != nil && r.SecurityGroup.Visibility != nil &&
+				*r.SecurityGroup.Visibility == "public" {
```
kobajagi, line 70: Although API might not return empty `SecurityGroup.Name`, it is still a pointer and adding ` && SecurityGroup.Name != nil` would not hurt.

PR #581, opened 2023-05-11
Add helper for public security groups

PhilippeChepy:
Implement missing bits for "public" security groups (e.g.: `public-nlb-healthcheck-sources`).


PhilippeChepy:
Not required for the decided implementation.
The CLI doesn't need it, and the terraform provider will rather expose a new "public_security_group" field whenever required.

PR #587, opened 2023-08-01
Egoscale v3 - base builder

kobajagi:


PhilippeChepy (approved):

jessicatoscani (approved):

kobajagi:
This PR is now main V3 development branch. All V3 PRs needs to merge into this one.

sauterp (approved):

v3/api/compute/compute.go
```diff
@@ -18,6 +18,10 @@ func NewCompute(c *oapi.ClientWithResponses) *Compute {
 //return NewInstances(a.oapiClient)
 //}
 
+func (a *Compute) InstanceTypess() *InstanceTypes {
```
sauterp, line 21:
```suggestion
func (a *Compute) InstanceTypes() *InstanceTypes {
```

kobajagi: Good catch.

v3/api/compute/instance_types.gen.go
```diff
@@ -0,0 +1,48 @@
...
+
+	err = utils.ParseResponseError(resp.StatusCode(), resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return *resp.JSON200.InstanceTypes, nil
+}
+
+func (a *InstanceTypes) Get(ctx context.Context, id openapi_types.UUID) (*oapi.InstanceType, error) {
```
7fELF, line 35: In my opinion we shouldn't use types from the generated oapi code.

The fact that we generate code from an OpenAPI spec behind the scene and then generate a wrapper on top is a technical detail for the user, and they shouldn't see it.

Having all the types in the api/** packages would make better docs and, more importantly it would keep the user facing API contained, allowing us to keep it stable even if the technical details behind the scene are changing (deepmap/oapi-codegen upgrades, or even replacement if it doesn't fit our needs or isn't maintained anymore).

(If we had custom generation, it would be different though, because we would have more control on the generated code)

v3/generator/generate.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"github.com/iancoleman/strcase"
+)
+
+// Generate is a command the generates Consumer API according to the mapping in mapping.go file.
+func Generate() {
+	tpl := template.Must(template.ParseFiles("templates/resource.tmpl"))
+
+	fset := token.NewFileSet()
+	node, err := parser.ParseFile(fset, "../oapi/oapi.gen.go", nil, parser.ParseComments)
```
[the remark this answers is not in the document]
kobajagi: Ideally yes, we would modify `deepmap/oapi-codegen` or even write code generator from scratch using parsed OpenAPI spec.

I think you underestimate complexity of `oapi-codegen` library. It allows us to use [custom templates](https://github.com/deepmap/oapi-codegen#making-changes-to-code-generation) which I tinkered with but I did not achieve much and would need to fork the library to implement everything i want. Learning curve is steep and I personally don't see it worth the time.

And although we have 2 layers of generated code, it is still generated code and maintenance is not an issue. Complexity of generator can be, from my POV it's simpler then what custom oapi-codegen would be.

v3/api/compute/instance_types.gen.go
kobajagi: Using types from oapi generated code is what makes generator very simple and wrapper thin. Copying structures to `api` namespace is doable, but i don't feel like it is worth the effort. I also don't think `oapi` code should be hidden, i see `api` and `oapi` as high level and low level api library that we offer.

sauterp: On this issue I tend to agree with @7fELF if we expose oapi types directly to users, we give up some control of our  API to `deepmap/oapi-codegen`. If we wrap the oapi types we will later be able to refuse changes that we don't agree with.
This even opens up the possibility of rewriting the code generation later. I still feel confident that our compromise of generating wrappers for `deepmap/oapi-codegen` instead of writing our own code generation tool makes sense. But if we make sure that the entire user-facing API is in our control, we keep the door open so that we may in the future rewrite the code generation process and remove the dependency on `deepmap/oapi-codegen`.

IIRC we are doing entire refactoring in part, because `deepmap/oapi-codegen` introduced breaking changes. IMHO avoiding this risk and having an API that we "fully" control justifies the extra effort of generating our own types.

v3/oapi/oapi.gen.go
brutasse, line 1: If this namespace is going to be exposed to users, could it be renamed to reflect that it exposes consts and structs for API calls payloads? The fact it comes from an OpenAPI spec is an implementation detail.

kobajagi:
@7fELF @brutasse I did a big refactor of this PR in regards to what we discussed.

Take a look and let me know if you have any further feedback.

v3/api/compute/instance_types.gen.go
kobajagi: This was partially addressed in the latest iteration. Types are generated in the same (root) package and are now in the separate file.

We do not copy generated types, doing so is non-trivial task with little advantage. Since types are now separated from codegen functionality we can replace them independently if needed (which boils down to the same problem as if we maintain a copy of types).

v3/client.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+			op, err := c.Global().Operations().Get(ctx, *op.Id)
+			if err != nil {
+				return nil, err
+			}
+			if *op.State != OperationStatePending {
```
pierre-emmanuelJ, line 135: Should be `==` here, no?

kobajagi:
Superseded by #602

PR #588, opened 2023-08-03
Egoscale V3 - Zone wrapper

kobajagi:
This PR adds zone-aware wrapper around V3 Client.

### DIFF
--- a/v3/zoned_client.go
+++ b/v3/zoned_client.go
@@ -28,7 +28,7 @@ var (
 )
 
 // ZonedClient is a Exoscale API Client that can communicate with API servers in different zones.
-// It has the same interface as Client and uses currently seleced zone to run API calls.
+// It has the same interface as Client and uses currently selected zone to run API calls.
 // Consumer is expected to select zone before invoking API calls.
 type ZonedClient struct {
 	zones       map[oapi.ZoneName]*oapi.ClientWithResponses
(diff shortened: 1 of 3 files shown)


PhilippeChepy (approved):

v3/zoned_client.go
```diff
@@ -0,0 +1,111 @@
...
+		oapi.AtVie2,
+		oapi.BgSof1,
+		oapi.ChDk2,
+		oapi.DeFra1,
+		oapi.DeMuc1,
+	}
+)
+
+// ZonedClient is a Exoscale API Client that can communicate with API servers in different zones.
+// It has the same interface as Client and uses currently seleced zone to run API calls.
```
jessicatoscani, line 31:
```suggestion
// It has the same interface as Client and uses currently selected zone to run API calls.
```

jessicatoscani (approved):

kobajagi:
Last commit adds support for env var overrides.

PhilippeChepy (approved):

v3/zoned_client.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+	DefaultHostPattern = "https://api-%s.exoscale.com/v2"
+
+	EnvKeyAPIEndpointPattern = "EXOSCALE_API_ENDPOINT_PATTERN"
+	EnvKeyAPIEndpointZones   = "EXOSCALE_API_ENDPOINT_ZONES"
+)
+
+var (
+	// DefaultZones list (available in oapi code).
+	// When new zone is added or existing removed this slice needs to be updated.
```
sauterp, line 22: I wish we could generate this somehow. Is there really no way?

sauterp (approved):

v3/zoned_client.go
kobajagi: We can generate it, it should be fairly straightforward. I'll create a separate story for that.

PR #589, opened 2023-08-07
V3: Consumer API

kobajagi:


v3/generator/templates/resource.tmpl
```diff
@@ -0,0 +1,35 @@
+package {{ .Package }}
+
```
sauterp, line 2: We could add a note like THIS CODE IS GENERATED or similar.

kobajagi: That's a good point and good practice. I'll add a note.

kobajagi:
```
// Code generated by v3/generator; DO NOT EDIT.
```

kobajagi: Done

PhilippeChepy (approved):

sauterp (approved):

PR #594, opened 2023-08-15
v3: ForEachZone helper and logging

kobajagi:
Example usage of `ForEachZone` and logging  with builtin `Logger` implementation (`StandardLogger`):

```go
func main() {
	client, err := exoscale.DefaultClient(
		exoscale.ClientOptWithCredentialsFromEnv(),
		exoscale.ClientOptWithLogger(exoscale.NewStandardLogger(os.Stdout, true)),
	)
	if err != nil {
		panic(err)
	}

	client.ForEachZone(func(c *exoscale.Client, zone oapi.ZoneName) {
		l, err := c.Compute().LoadBalancers().Get(context.Background(), <id>)
		if err != nil {
			fmt.Println(err)
		} else {
			fmt.Println(l)
		}
	})
}
```

### DIFF
--- a/v3/logger.go
+++ b/v3/logger.go
@@ -28,27 +28,27 @@ func NewStandardLogger(writer io.Writer, debug bool) *StandardLogger {
 }
 
 func (l *StandardLogger) Error(msg string, keysAndValues ...interface{}) {
-	fmt.Fprintf(l.writer, "[ERROR] %s %v\n", msg, toMap(keysAndValues))
+	fmt.Fprintf(l.writer, "[ERROR] %s %v\n", msg, l.toMap(keysAndValues))
 }
 
 func (l *StandardLogger) Info(msg string, keysAndValues ...interface{}) {
-	fmt.Fprintf(l.writer, "[INFO] %s %v\n", msg, toMap(keysAndValues))
+	fmt.Fprintf(l.writer, "[INFO] %s %v\n", msg, l.toMap(keysAndValues))
 }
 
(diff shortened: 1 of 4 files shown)


PhilippeChepy (approved):

v3/logger.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil
+	}
+
+	if len(keysAndValues)%2 > 0 {
+		keysAndValues = append(keysAndValues, nil)
+	}
+
+	m := map[string]interface{}{}
+	for i := 0; i < len(keysAndValues); i = i + 2 {
+		if k, ok := keysAndValues[i].(string); ok {
```
sauterp, line 62: IIUC this will silently ignore a pair of key and value if the key is not a string. Could we log an error or warning in that case? If someone accidentally doesn't specify arguments in the right order it won't log them and the user might not notice.

kobajagi:
> Could it be useful to have an optional argument for ForEachZone that allows users to specify a subset of the zones?

Idk if that is the usecase that justifies extra code. Helpers should solve common problems, zones subset iteration doesn't sound common (i cannot think of any usage internally). I'd left it for consumers to implemen.

v3/logger.go
kobajagi: Good point. And it's convenient that function is called by the logger itself so i have log writer ready.

kobajagi: Fixed

sauterp (approved):

PR #610, opened 2023-12-13
Egoscale V3: Replace Zone with url endpoints

pierre-emmanuelJ:
# Description

Add zoned URL to replace standalone zones and useless `Environment`

### DIFF
--- a/v3/client.go
+++ b/v3/client.go
@@ -28,8 +28,8 @@ const (
 	BGSof1 URL = "https://api-bg-sof-1.exoscale.com/v2"
 )
 
-// URLs represents a list of zoned URL.
-var URLs map[string]URL = map[string]URL{
+// Zones represents a list of all Exoscale Zone.
+var Zones map[string]URL = map[string]URL{
 	"ch-gva-2": CHGva2,
 	"ch-dk-2":  CHDk2,
 	"de-fra-1": DEFra1,
--- a/v3/client.go
+++ b/v3/client.go
@@ -28,7 +28,7 @@ const (
 	BGSof1 URL = "https://api-bg-sof-1.exoscale.com/v2"
 )
 
-// Zones represents a list of all Exoscale Zone.
+// Zones represents a list of all Exoscale zone.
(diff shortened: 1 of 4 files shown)


v3/client.go
```diff
@@ -9,30 +9,41 @@ import (
...
+	CHDk2  URL = "https://api-ch-dk-2.exoscale.com/v2"
+	DEFra1 URL = "https://api-de-fra-1.exoscale.com/v2"
+	DEMuc1 URL = "https://api-de-muc-1.exoscale.com/v2"
+	ATVie1 URL = "https://api-at-vie-1.exoscale.com/v2"
+	ATVie2 URL = "https://api-at-vie-2.exoscale.com/v2"
+	BGSof1 URL = "https://api-bg-sof-1.exoscale.com/v2"
 )
 
+// URLs represents a list of zoned URL.
+var URLs map[string]URL = map[string]URL{
```
aureliar8, line 32:
```suggestion
var Zones map[string]URL = map[string]URL{
```
Maybe this is clearer for the variable name

aureliar8 (approved):

kobajagi (approved):

v3/client.go
```diff
@@ -113,7 +117,6 @@ func NewClient(apiKey, apiSecret string, opts ...ClientOpt) (*Client, error) {
 		apiKey:          apiKey,
 		apiSecret:       apiSecret,
 		serverURL:       "https://api-ch-gva-2.exoscale.com/v2",
```
sauterp, line 119:
```suggestion
		serverURL:       string(CHGva2),
```
didn't test this

sauterp (approved):

v3/client.go
pierre-emmanuelJ: I need to modify the generator, to change the default constructor logic, not the generated file. Let me try

pierre-emmanuelJ: https://github.com/exoscale/egoscale/pull/610/commits/f2289d326c5869140b956050a4d2f74fe5ea35be

PR #612, opened 2023-12-15
V3: Add Credentials provider

pierre-emmanuelJ:
# Description
Add credentials provider to manages Exoscale secret from different environment and time.

## Checklist
(For exoscale contributors)

V3 alpha development


v3/credentials.go
```diff
@@ -0,0 +1,156 @@
...
+	c.injectors = append(c.injectors, f...)
+}
+
+var InjectorFile InjectorFn = func(ctx context.Context) (CredentialsValue, error) {
+	return CredentialsValue{}, nil
+}
+
+// InjectorEnv retrieve credentials from ENV variables.
+// InjectorEnv credentials never expire.
+var InjectorEnv InjectorFn = func(ctx context.Context) (CredentialsValue, error) {
```
sauterp, line 115: Since this is in the v3 package we could put "Credentials" in the name, to make clear what the injector is about. Suggestions: "EnvCredentialsInjector", "GetEnvCredentials", "InjectEnvironmentCredentials".

v3/credentials.go
```diff
@@ -0,0 +1,156 @@
...
+		return CredentialsValue{}, fmt.Errorf("missing or incomplete API credentials")
+	}
+
+	return CredentialsValue{
+		APIKey:    apiKey,
+		APISecret: apiSecret,
+	}, nil
+}
+
+func Test() {
```
sauterp, line 129: Why not put it in a _test.go file? ;)

v3/credentials.go
aureliar8: I'm also wondering if we should make this a one method interface rather than a function.
That way providing a custom implementation that relies on some other state may be easier than trying to build the correct closure.

v3/credentials.go
pierre-emmanuelJ: Just to show the draft usage as example here, but yes let's add it in a test file once the design is final and stable 👍

v3/credentials/chain.go
```diff
@@ -0,0 +1,57 @@
...
+	return NewCredentials(&ChainProvider{
+		Providers: append([]Provider{}, providers...),
+	})
+}
+
+// Retrieve returns the credentials value or error if no provider returned
+// without error.
+//
+// If a provider is found it will be cached and any calls to IsExpired()
+// will return the expired state of the cached provider.
```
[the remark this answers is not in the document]
kobajagi: Also clarify that function returns value returned by the first provider in the chain that succeeds.

kobajagi (changes requested):

v3/credentials/env.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Retrieve retrieves the keys from the environment.
+func (e *EnvProvider) Retrieve() (Value, error) {
+	e.retrieved = false
+
+	v := Value{
+		APIKey: readFromEnv(
+			"EXOSCALE_API_KEY",
+			"EXOSCALE_KEY",
```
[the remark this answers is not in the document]
pierre-emmanuelJ: https://github.com/exoscale/egoscale/pull/612/commits/27ac8e6364f1d46edadd56c57dac19e4959683bd

kobajagi (approved):

pierre-emmanuelJ:
Again dealing with this HashiCorp vendor CRLF issue.. Causing the CI to fail

v3/credentials/chain.go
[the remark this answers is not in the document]
pierre-emmanuelJ: Thanks, I'm adding more context to the comment.

Concerning the concurrency,
All provider implementation are not concurrent safe itself (env, chain, file...etc).
It's the usage of providers with credentials are protected here. 
Credentials is the main logic (retrieve, mutex...etc) and a provider is like a Driver.

IMHO, if we specify this one is not concurrent safe, we need to notice for all the provider implementation then.

sauterp (approved):

PR #615, opened 2024-01-23
V3: Add better message on validator

pierre-emmanuelJ:
# Description
More explicit error message on the default validator

## Checklist
(For exoscale contributors)

V3 alpha development

### DIFF
--- a/v3/api.go
+++ b/v3/api.go
@@ -127,8 +127,31 @@ func Ptr[T any](v T) *T {
 }
 
 // Validate any struct from schema or request
-func Validate(r any) error {
-	return validator.New().Struct(r)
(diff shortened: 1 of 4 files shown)


aureliar8 (approved):
Do you have an example of the new message for the same error ?

wdullaer (approved):

pierre-emmanuelJ:
```Golang
2024/01/24 09:20:29 Request validation error: 'CreateBlockStorageVolumeRequest.Size' = '99' does not validate 'gte=100'
```
What do you think about it @aureliar8 and @wdullaer? Is it more explicit
maybe the `gte` but it's understandable enough to me

PR #618, opened 2024-02-22
go.mk: remove submodule and initialize through make

sauterp:
## Checklist
(For exoscale contributors)

* [x] Changelog updated (under *Unreleased* block)
* [ ] For a new resource or new attributes: test added/updated

PR #619, opened 2024-03-07
Makefile: fix targets

sauterp:
# Description
Due to the introduction of `.ONESHELL` to `Makefile` in #618  the shell commands in each target are now executed in one shell session. This improves performance but also means that changing the directory in one line also changes the directory for the following lines(doesn't happen across targets). Unfortunately it broke the targets `oapigen` and `generate`.

## Checklist
(For exoscale contributors)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,8 +71,10 @@ oapigen: install-oapi-codegen
 .PHONY: generate
 generate:
 	@wget -q --show-progress --progress=dot https://openapi-v2.exoscale.com/source.yaml -O- > v3/generator/source.yaml
-	@echo
-	@cd v3/generator/; go generate
-	@go mod tidy && go mod vendor
-	@rm v3/generator/source.yaml
-	@ls -l v3/*.go
+	@cd v3/generator/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,17 +62,19 @@ install-oapi-codegen:
...
 	@wget -q --show-progress --progress=dot https://openapi-v2.exoscale.com/source.yaml -O- > v3/generator/source.yaml
-	@echo
-	@cd v3/generator/; go generate
-	@go mod tidy && go mod vendor
-	@rm v3/generator/source.yaml
-	@ls -l v3/*.go
+	@cd v3/generator/
+	@go generate
+	@go mod tidy
+	@go mod vendor
```
pierre-emmanuelJ, line 77: This need to be executed a the golang root repo?

pierre-emmanuelJ: Perfect! Fair enough 👍

pierre-emmanuelJ (approved):

kobajagi (approved):

PR #622, opened 2024-04-05
v3 trace: print operation IDs

sauterp:
# Description
To identify which API operations a given call executes we print the operation ID in addition to the request dump when tracing is enabled. For a CLI command the output will look like this:
```shell
❯ EXOSCALE_TRACE=1 exo compute -z ch-gva-2 bs list
>>> Operation: list-zones
GET /v2/zone HTTP/1.1
Host: api-ch-gva-2.exoscale.com

...

>>> Operation: list-block-storage-volumes
GET /v2/block-storage HTTP/1.1
Host: api-ch-gva-2.exoscale.com
```

## Checklist
(For exoscale contributors)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Unreleased
 - go.mk: remove submodule and initialize through make #618 
 - Makefile: fix targets #619 
 - SKS nodepool: support for kubelet image gc #620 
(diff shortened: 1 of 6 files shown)


alexanderjamesking (approved):

arnaudgeiser (approved):
Thanks!

PR #628, opened 2024-04-24
V3: Remove omitempty on nullable complex types

pierre-emmanuelJ:
# Description
Update instance pool definition
```Yaml
    put:
      tags:
      - instance-pool
      responses:
        '200':
          description: '200'
          content:
            application/json:
              schema:
                "$ref": "#/components/schemas/operation"
      description: ''
      parameters:
      - in: path
        required: true
        name: id
        schema:
          type: string
          format: uuid
      summary: Update an Instance Pool
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                anti-affinity-groups:
                  type: array
                  items:
                    "$ref": "#/components/schemas/anti-affinity-group"
                  uniqueItems: true
                  nullable: true
                  description: Instance Pool Anti-affinity Groups
                description:
                  type: string
                  maxLength: 255
                  description: Instance Pool description
                public-ip-assignment:
                  "$ref": "#/components/schemas/public-ip-assignment"
                  description: Assign public IP to Instances
                labels:
                  "$ref": "#/components/schemas/labels"
                  description: Instance Pool Labels
                security-groups:
                  type: array
                  items:
                    "$ref": "#/components/schemas/security-group"
                  uniqueItems: true
                  nullable: true
                  description: Instance Pool Security Groups
                elastic-ips:
                  type: array
                  items:
                    "$ref": "#/components/schemas/elastic-ip"
                  nullable: true
                  description: Instances Elastic IPs
                name:
                  type: string
                  maxLength: 255
                  minLength: 1
                  description: Instance Pool name
                instance-type:
                  "$ref": "#/components/schemas/instance-type"
                  description: Instances type
                min-available:
                  type: integer
                  format: int64
                  minimum: 0
                  exclusiveMinimum: false
                  nullable: true
                  description: Minimum number of running Instances
                private-networks:
                  type: array
                  items:
                    "$ref": "#/components/schemas/private-network"
                  uniqueItems: true
                  nullable: true
                  description: Instance Pool Private Networks
                template:
                  "$ref": "#/components/schemas/template"
                  description: Instances template
                ssh-key:
                  "$ref": "#/components/schemas/ssh-key"
                  nullable: true
                  description: Instances SSH key
                instance-prefix:
                  type: string
                  nullable: true
                  description: 'Prefix to apply to Instances names (default: pool)'
                user-data:
                  type: string
                  minLength: 1
                  nullable: true
                  description: Instances Cloud-init user-data
                deploy-target:
                  "$ref": "#/components/schemas/deploy-target"
                  nullable: true
                  description: Instance Pool Deploy Target
                ipv6-enabled:
                  type: boolean
                  description: 'Enable IPv6. DEPRECATED: use `public-ip-assignments`.'
                disk-size:
                  type: integer
                  format: int64
                  minimum: 10
                  maximum: 50000
                  exclusiveMinimum: false
                  exclusiveMaximum: false
                  description: Instances disk size in GB
                ssh-keys:
                  type: array
                  items:
                    "$ref": "#/components/schemas/ssh-key"
                  uniqueItems: true
                  nullable: true
                  description: Instances SSH keys
      operationId: update-instance-pool
```

Probably some missing `nullable` statement is missing in this spec for `e.g: labels`

## Checklist
(For exoscale contributors)

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -332,11 +332,12 @@ func renderObject(typeName string, s *base.Schema, output *bytes.Buffer) (string
 		tag := fmt.Sprintf(" `json:\"%s,omitempty\"", propName)
 
 		pointer := "*"
-		req := isRequiredField(propName, s)
-		if req || nullable {
+		required := isRequiredField(propName, s)
+		// Do not remove omitempty on simple nullable types.
+		if required || (nullable && !IsSimpleSchema(prop)) {
 			tag = fmt.Sprintf(" `json:%q", propName)
 		}
-		validation := renderValidation(prop, req)
+		validation := renderValidation(prop, required)
 		if validation != "" {
 			tag += " " + validation
 		}
@@ -493,7 +494,7 @@ func IsSimpleSchema(s *base.Schema) bool {
 		return true
 	}
 
-	return s.Type[0] != "object" && s.Type[0] != "map"
+	return s.Type[0] != "object" && s.Type[0] != "map" && s.Type[0] != "array"
 }
 
 func renderDoc(s *base.Schema) string {
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -333,7 +333,7 @@ func renderObject(typeName string, s *base.Schema, output *bytes.Buffer) (string
 
 		pointer := "*"
 		req := isRequiredField(propName, s)
-		if req {
+		if req || nullable {
 			tag = fmt.Sprintf(" `json:%q", propName)
 		}
 		validation := renderValidation(prop, req)
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -486,9 +486,7 @@ func InferType(s *base.Schema) {
 	}
 }
 
-// IsSimpleSchema returns whether this schema is a scalar or array as these
-// can't be circular references. Objects result in `false` and that triggers
-// circular ref checks.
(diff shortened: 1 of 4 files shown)


kobajagi (approved):

v3/generator/schemas/schemas.go
```diff
@@ -493,7 +494,7 @@ func IsSimpleSchema(s *base.Schema) bool {
 		return true
 	}
 
-	return s.Type[0] != "object" && s.Type[0] != "map"
+	return s.Type[0] != "object" && s.Type[0] != "map" && s.Type[0] != "array"
```
sauterp, line 497: The comment of `IsSimpleSchema` indicates that you explicitly didn't add "array" here because it can't have circular references. It sounds like this function has multiple purposes now.

Why not create a function calles `IsSimpleSchema` so we don't mix the two purposes.
If not we should change the comment of this function.

sauterp (changes requested):

v3/generator/schemas/schemas.go
pierre-emmanuelJ: Thanks @sauterp let me check now the usage of this `func` in the codebase and let's see what's fits the best

pierre-emmanuelJ: After checking every call, I refactored the comment, and also the usage is better like that, even in the rest of the codebase.
This comment was referencing a logic before I refactored some stuff in the generator

sauterp (approved):
thanks~

PR #634, opened 2024-06-12
V3 meta-data: private Instance fetch metadata from CD-ROM

pierre-emmanuelJ:
# Description
This change is useful in the context of working with Exoscale private instance metadata (useful for CCM, CSI...etc).

In the CSI container are run as privileged by default to access device for block storage...etc., for CCM the manual prerequisite of the customer if he installs it on his private instance cluster is to change the container permission

## Checklist
(For exoscale contributors)

Tested from private instance with a test program 

```
ubuntu@test-private:~$ sudo ./test-meta
Response for availability-zone: ch-gva-2
Response for cloud-identifier: CloudStack-{2f8551a8-xxxx7ec69527fb}
Response for instance-id: fd1c0d3c-xxxxxb1d5dda
Response for local-hostname: test-private
Response for public-hostname: test-private
Response for service-offering: Medium 4096mb 2cpu
Response for vm-id: fd1c0d3c-xxxxxb1d5dda
```

if no sudo executed

```
ubuntu@test-private:~$ ./test-meta
2024/06/12 07:56:14 disk open: could not open device /dev/disk/by-label/cidata with mode 0: open /dev/disk/by-label/cidata: permission denied
....etc
```

if not, a private instance
```
ubuntu@bastion:~$ ./test-meta
2024/06/12 07:57:00 disk open: provided device /dev/disk/by-label/cidata does not exist
Response for availability-zone:
```


Fumesover (commented):
It feels weird to mount & unmount the disk each time you need to access some metadata. And since you're using a hardcoded target, two processes could mount it at the same time and cause some issues

Did you consider the possibility of reading straight from the disk instead of mounting it on the system? I didn't looked too much into it, but it looks like https://github.com/diskfs/go-diskfs might do the trick

pierre-emmanuelJ:
> It feels weird to mount & unmount the disk each time you need to access some metadata. And since you're using a hardcoded target, two processes could mount it at the same time and cause some issues
> 
> Did you consider the possibility of reading straight from the disk instead of mounting it on the system? I didn't looked too much into it, but it looks like https://github.com/diskfs/go-diskfs might do the trick

@Fumesover 

Here is in my new commit a better solution as you proposed, what do you think about?

v3/metadata/metadata.go
```diff
@@ -70,3 +92,20 @@ func httpGet(ctx context.Context, url string) (string, error) {
 
 	return string(body), nil
 }
+
+func getFileMetaDataValue(file *os.File, endpoint string) (string, error) {
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
```
Fumesover, line 98: It feels weird to read the raw disk & search for a string inside of it like that

Did you consider using `Disk.GetFilesystem` then `FileSystem.OpenFile` ?

v3/metadata/metadata.go
```diff
@@ -66,26 +64,13 @@ func Get(ctx context.Context, endpoint Endpoint) (string, error) {
...
-
-	if err := unix.Mount(CdRomPath, target, "iso9660", unix.MS_RDONLY, ""); err != nil {
-		return "", fmt.Errorf("mount %s: %w", CdRomPath, err)
+		return "", fmt.Errorf("disk open: %w", err)
 	}
-	defer unix.Unmount(target, 0) //nolint: errcheck
+	defer iso.File.Close()
 
-	return getFileMetaDataValue(filepath.Join(target, "meta-data"), string(endpoint))
+	return getFileMetaDataValue(iso.File, string(endpoint))
```
aureliar8, line 73: If the only thing you do is just to reuse the underlying file descriptor to read it, can't you just `os.Open()` directly without the dependency on diskfs ?  

Unless I missed something that the lib does

pierre-emmanuelJ: I will try to use and make work:
 using `Disk.GetFilesystem` then `FileSystem.OpenFile` ?
Otherwise, yes if it's not working I can use open directly

v3/metadata/metadata.go
```diff
@@ -51,6 +59,34 @@ func Get(ctx context.Context, endpoint Endpoint) (string, error) {
...
+// Important note: Run this code as privileged user.
+// Not Windows compatible.
+func FromCdRom(endpoint Endpoint) (string, error) {
+	disk, err := diskfs.Open(CdRomPath, diskfs.WithOpenMode(diskfs.ReadOnly))
+	if err != nil {
+		return "", fmt.Errorf("disk open: %w", err)
+	}
+
+	// TODO: Fix the block size in orchestrator from 512 to 2048
+	disk.DefaultBlocks = true
```
aureliar8, line 73: Don't you need to close the disk at one point ?

pierre-emmanuelJ: let me check with the diskfs lib thanks

pierre-emmanuelJ: Tested successfully https://github.com/exoscale/egoscale/pull/634/commits/cddde218eb353ae2970f34d7ee98b583292c95f3 it the only way to close it, and it works

Fumesover (approved):

v3/metadata/metadata.go
```diff
@@ -70,3 +106,20 @@ func httpGet(ctx context.Context, url string) (string, error) {
...
+		if len(parts) != 2 {
+			continue
+		}
+
+		if strings.TrimSpace(parts[0]) == endpoint {
+			return strings.TrimSpace(parts[1]), nil
+		}
+	}
+
+	return "", fmt.Errorf("endpoint '%s' not found in file", endpoint)
```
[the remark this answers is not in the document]
pierre-emmanuelJ: https://github.com/exoscale/egoscale/pull/634/commits/f48388592b7ba3415f3fba67215751bf0f84cd61

Fumesover (approved):

kobajagi (approved):

PR #641, opened 2024-07-11
V3: Add more HTTP error types

pierre-emmanuelJ:
# Description
Add more HTTP error types to be more efficient on our api errors

## Checklist
(For exoscale contributors)

Test:

```Go
func main() {
	c := credentials.NewEnvCredentials()
	cli, err := v3.NewClient(c)
	if err != nil {
		log.Fatal(err)
	}

	_, err = cli.GetInstance(context.Background(), "bad request not an uuid")
	if err != nil {
		if errors.Is(err, v3.ErrBadRequest) {
			log.Println("OK")
		}
	}

	_, err = cli.GetInstance(context.Background(), "e4a96b3f-9d50-4137-bb02-ba734c75a886")
	if err != nil {
		if errors.Is(err, v3.ErrNotFound) {
			log.Println("OK")
		}
	}
}

```

Output:

```
2024/07/11 16:47:39 OK
2024/07/11 16:47:39 OK
```

We will need to update CSI:
https://github.com/exoscale/exoscale-csi-driver/blob/main/driver/controller.go#L337
from `ErrInvalidRequest` to `ErrBadRequest` (the dedicated code to the returned error)


go.mod
```diff
@@ -8,14 +8,12 @@ require (
...
-require (
-	github.com/kr/pretty v0.3.1 // indirect
-	gopkg.in/yaml.v3 v3.0.1 // indirect
-)
+require gopkg.in/yaml.v3 v3.0.1 // indirect
 
 require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.2 // indirect
+	github.com/kr/pretty v0.3.1 // indirect
```
sauterp, line 16: the changes in this file look like noise to me, how did they come about?

v3/errors.go
```diff
@@ -0,0 +1,127 @@
...
+			if err = json.Unmarshal(data, &res); err != nil {
+				return fmt.Errorf("error unmarshaling response: %s", err)
+			}
+		} else {
+			res.Message = string(data)
+		}
+
+		err, ok := httpStatusCodeErrors[resp.StatusCode]
+		if ok {
+			if err == ErrNotFound {
```
sauterp, line 117: Why do we need this special case?

sauterp (approved):

v3/errors.go
pierre-emmanuelJ: I reproduced the past behavior, to not ship the message (probably no message for 404 that why) in the error for 404 but I think I can test to do like other codes, yes

pierre-emmanuelJ: I removed it message is here and more explicit then, thanks @sauterp

v3/errors.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+var (
+	ErrBadRequest                    = fmt.Errorf(http.StatusText(http.StatusBadRequest))
```
aureliar8, line 11:
```suggestion
	ErrBadRequest                    = errors.New((http.StatusText(http.StatusBadRequest))
```

pierre-emmanuelJ: https://github.com/exoscale/egoscale/pull/641/commits/4a5dab3dafa22d32955f76dae394378f17177e2e

aureliar8 (approved):

aureliar8 (commented):
Just note that it's a breaking change. 

I'm not sure what's the current policy for egoscale v3

PR #643, opened 2024-07-16
v3: automate regeneration when spec changes

sauterp:
# Description
Everyday at 7 AM we trigger a GitHub Action that runs `make generate`. This will download the exoscale API spec and try to regenerate egoscale v3. If there are changes, we create a PR with those changes.

NOTE: the latest run of the action fails, because I changed the base to `master` which means there's a conflict, because source.yaml is ignored in `master`. [example](https://github.com/exoscale/egoscale/actions/runs/9955412776/job/27503113843?pr=643#step:4:108)
Once this PR is merged the problem will disappear. See testing below for tests on a previous commit.

## Checklist
(For exoscale contributors)

## Testing
Pushing this PR automatically [generated this PR](https://github.com/exoscale/egoscale/pull/644)

### DIFF
--- a/.github/workflows/generate_v3.yaml
+++ b/.github/workflows/generate_v3.yaml
@@ -1,4 +1,4 @@
-name: Check whether the exoscale API has been updated and regenerate egoscale v3.
+name: check_and_regenerate_v3
 
 # TODO run on a regular basis
 on:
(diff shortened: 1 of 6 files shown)


arnaudgeiser (approved):
Awesome!

PhilippeChepy (approved):
Cool stuff

PR #656, opened 2024-09-11
Generator: Add Custom Extension Findable

pierre-emmanuelJ:
# Description

From the spec findable `x-go-findable` `1` and `2` it's building the custom findable we want.
By default, it's remain `name 1`  and `id 2` but now you can for example give

for elastic-ip you can point:
`1 - id`
`2 - ip`

and findable will generate a findable func with `idOrIP` parameter to be found 

## Checklist
(For exoscale contributors)

### DIFF
--- a/v3/README.md
+++ b/v3/README.md
@@ -75,11 +75,11 @@ func main() {
 Most of the list request `ListX()` return a type containing the list of the resource requested and a method `FindX()` to be able to retrieve a resource by its `name` or `id` most of the time.
 
 ```Golang
-instances, err := client.ListInstancePools(ctx)
+pools, err := client.ListInstancePools(ctx)
 if err != nil {
 	log.Fatal(err)
 }
-pool, err := instances.FindInstancePool("my-pool-example")
+pool, err := pools.FindInstancePool("my-pool-example")
 if err != nil {
 	log.Fatal(err)
 }


elkezza (approved):
lgtm :)

v3/README.md
```diff
@@ -70,6 +70,23 @@ func main() {
...
+### Findable
+
+Most of the list request `ListX()` return a type containing the list of the resource requested and a method `FindX()` to be able to retrieve a resource by its `name` or `id` most of the time.
+
+```Golang
+instances, err := client.ListInstancePools(ctx)
+if err != nil {
+	log.Fatal(err)
+}
+pool, err := instances.FindInstancePool("my-pool-example")
```
kobajagi, line 82:
```suggestion
pools, err := client.ListInstancePools(ctx)
if err != nil {
	log.Fatal(err)
}
pool, err := pools.FindInstancePool("my-pool-example")
```

v3/README.md
```diff
@@ -70,6 +70,23 @@ func main() {
 }	
 ```
 
+### Findable
+
+Most of the list request `ListX()` return a type containing the list of the resource requested and a method `FindX()` to be able to retrieve a resource by its `name` or `id` most of the time.
```
kobajagi, line 75: `X` in `FindX()` is not really needed. All functions could just be named `Find()` as they are method of specific struct.
It's a breaking change though, you can decide if it's worth changing or leave it as is.

pierre-emmanuelJ: For the moment we will leave as it is to not breaking change....

v3/README.md
```diff
@@ -87,3 +104,53 @@ GENERATOR_DEBUG=client make generate > test/client.go
...
+	Generated code
+	```Golang
+	type Endpoint string
+
+	type Zone struct {
+		APIEndpoint Endpoint // Here is the generated type definition.
+		...
+	}
+	```
+- `x-go-findable` to specify which fields in the findable resource to fetch
```
kobajagi, line 129: From my understanding of the code, there are some limitations worth documenting:
- `x-go-findable` can only be set on string types;
- max 2 attributes can have a `x-go-findable`;
- values of `x-go-findable` can be either `"1"` or `"2"` , where former value determines string will be search there first.

kobajagi (approved):

PR #659, opened 2024-09-12
v3: user agent as client options

ldez:
# Description

Currently, the user agent is defined by a global variable, but this is a problem.

If lego defines the user agent and an app uses exocale and lego, what will be the user agent at the end?
It's not possible to know because it will depend on when and where the user agent is set:
- If lego is the last to set the variable then all the calls will have the lego user agent.
- If lego is the first to set the variable then all the calls will have the app user agent.
- If the app and/or lego defined the user agent through another global variable, it will be "random".

Another problem: with the global variable, if the user defines the user agent then information about the API client (versions, etc.) will be discarded, except if the user appends the previous user agent manually.

Generally speaking, global variables should be avoided for a lib.

---

I added a `ClientOpt` implementation to set the user agent: (`ClientUserAgent()`) and deprecated the global variable.

I created a compatibility layer, so the usage of the global variable still works:
- for a client, if the global variable `UserAgent` is overridden and `ClientUserAgent()` is not used, the user agent will be `UserAgent`.
- for a client, if `ClientUserAgent()` is used, even if `UserAgent` is overridden, the user agent will be the one defined by `ClientUserAgent()`

2 clients can use either the `UserAgent` or `ClientUserAgent()`, and each client will not be impacted by the other (except if both use the global variable as before).


v3/generator/client/client.tmpl
```diff
@@ -134,6 +149,21 @@ func NewClient(credentials *credentials.Credentials, opts ...ClientOpt) (*Client
...
+
+	if UserAgent != defaultUA {
+		return UserAgent
+	}
+
+	return c.userAgent
+}
+
 // WithEndpoint returns a copy of Client with new zone Endpoint.
 func (c *Client) WithEndpoint(endpoint Endpoint) *Client {
```
pierre-emmanuelJ, line 168: Thanks! Can you also update all the `WithX` to make sure user agent is copied?
Feel free to add a:
```Golang
func (c *Client) WithUserAgent(userAgent string) *Client {
```

`WithX` method are useful when we want to return a copy of the client with a new option
```Golang
client2 := client.WithUserAgent("test user-agent").WithEnpoint(x).....
```

v3/generator/client/client.tmpl
```diff
@@ -199,3 +218,17 @@ func (c *Client) executeRequestInterceptors(ctx context.Context, req *http.Reque
 
 	return nil
 }
+
+func cloneClient(c *Client) *Client {
```
pierre-emmanuelJ, line 222: Thanks for the improvement :)

pierre-emmanuelJ (approved):

sauterp (approved):
Thanks for your work! LGTM
I added one small rename to align the naming with the other client options we have.

sauterp:
tested the change in the cli:
`User-Agent: exoscale-cli/dev egoscale/v3.1.4 (go1.23.0; linux/amd64)`

(the dev will disappear once we do a cli release)

aureliar8 (changes requested):
Unless I'm mistaken @sauterp you only modified the client.go and not the code that generates it ?

sauterp:
thanks @aureliar8 !
I fixed it now

aureliar8 (approved):

PR #671, opened 2024-11-14
Generator: Fix additional properties spec 3.1

pierre-emmanuelJ:
# Description

## Checklist
(For exoscale contributors)

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -205,10 +205,6 @@ func renderSchemaInternal(schemaName string, s *base.Schema, output *bytes.Buffe
 
 func renderSimpleTypeEnum(typeName string, s *base.Schema) string {
 	for _, e := range s.Enum {
-		if s.Minimum != nil || s.Maximum != nil {
-			return ""
-		}
-
 		if strings.Contains(e.Value, ",") {
 			return ""
 		}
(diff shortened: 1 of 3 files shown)


kobajagi (approved):

pierre-emmanuelJ:
It why I did this dirty humanize integer workaround:
image
to prevent name collision....
And also if the value is not alphanumeric, I print the index of the enum instead of the value in the const name suffix

that's two workaround to make the spec compile

v3/generator/schemas/schemas.go
```diff
@@ -526,3 +537,22 @@ func isNullableReference(node *yaml.Node) bool {
...
+	if err != nil {
+		return s
+	}
+
+	if n >= 0 {
+		return s
+	}
+
+	// remove the "-" with helper
+	return helpers.ToLowerCamel(fmt.Sprintf("minus%s", s))
```
sauterp, line 557: why is ToLowerCamel still needed here?

pierre-emmanuelJ: it's dropping the char "-" to make it lower camel: "minus-1" become "minus1"

kobajagi (approved):

v3/generator/schemas/schemas.go
```diff
@@ -203,16 +204,32 @@ func renderSchemaInternal(schemaName string, s *base.Schema, output *bytes.Buffe
 }
 
 func renderSimpleTypeEnum(typeName string, s *base.Schema) string {
+	for _, e := range s.Enum {
+		if s.Minimum != nil || s.Maximum != nil {
+			return ""
+		}
+
```
sauterp, line 211: When the new public-api with the sanitized spec is published, you may want to delete these lines again and regenerate before merging.

sauterp (approved):

PR #710, opened 2025-07-23
v3: use go-retryabble as default HTTP client

kobajagi:
# Description

Copies the implementation from v2.
Default retry policy logic is implemented [here](https://github.com/hashicorp/go-retryablehttp/blob/v0.7.8/client.go#L495).

## Checklist
(For exoscale contributors)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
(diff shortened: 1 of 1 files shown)


pierre-emmanuelJ (approved):
LGTM

PR #762, opened 2026-04-23
v3: read RFC 9457 detail and title from error responses

loicrouillermonay:
# Description
v3: handleHTTPErrorResp now reads detail and title fields from RFC 9457 problem+json error responses.

Some APIs return errors in RFC 9457 format with detail and title fields. The Go client only looked for message and error fields. Both are absent in RFC 9457 responses, so the error message stayed empty. Users saw "Bad Request:" with nothing after the colon.

I added Detail and Title to the response struct and extended the fallback chain to message → error → detail → title. Legacy responses with message/error fields still work. RFC 9457 responses now surface their error details.

## Checklist
(For exoscale contributors)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+unreleased
+----------
+
+- v3: read RFC 9457 detail and title from error responses
+
 3.1.35
 ----------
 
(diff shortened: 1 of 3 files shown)


habfast:
Replaced by #763

PR #763, opened 2026-04-23
improved error management for RFC-9457 error responses

habfast:
# Description

In ADR006 in public-api, errors should be compliant with RFC-9457:
https://github.com/exoscale/public-api/blob/be90055610687fc3088a0cc8365479ca33cddb46/doc/adr/0006-openapi-spec-conventions.md?plain=1#L350

These aren't well supported in the CLI yet:
Screenshot 2026-04-23 at 13 09 58

With this PR, we can improve this quite a bit (PR to come in the CLI):
Screenshot 2026-04-23 at 13 24 14

Shortcut ticket: [sc-176689](https://app.shortcut.com/exoscale/story/176689/improve-error-management-of-cli)

## Checklist
(For exoscale contributors)

### DIFF
--- a/v3/errors.go
+++ b/v3/errors.go
@@ -94,11 +94,41 @@ var httpStatusCodeErrors = map[int]error{
 	http.StatusNetworkAuthenticationRequired: ErrNetworkAuthenticationRequired,
 }
 
+// APIError is the structured error returned for HTTP 4xx/5xx responses.
+// It wraps the HTTP sentinel error (e.g. ErrNotFound) so that errors.Is
+// checks against sentinels continue to work unchanged.
+type APIError struct {
+	// sentinel is the mapped HTTP status error (e.g. ErrNotFound).
+	sentinel error
+	// Message is the plain server message (from "message" or "error" fields).
+	Message string
+	// Title is the RFC 9457 "title" field.
+	Title string
+	// Detail is the RFC 9457 "detail" field.
+	Detail string
+	// Errors holds the RFC 9457 "errors" entries.
+	Errors []APIErrorEntry
+}
+
+// APIErrorEntry is one item from the RFC 9457 "errors" array.
+// Location is a JSONPath expression (e.g. "$['some-field']") when the server
+// provides field-level detail; it is empty for plain string error entries.
+type APIErrorEntry struct {
+	Detail   string
+	Location string
+}
+
+func (e *APIError) Error() string { return e.sentinel.Error() }
+func (e *APIError) Unwrap() error { return e.sentinel }
+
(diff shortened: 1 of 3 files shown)


v3/errors.go
```diff
@@ -129,3 +178,40 @@ func handleHTTPErrorResp(resp *http.Response) error {
 
 	return nil
 }
+
+// formatProblemErrors formats the RFC 7807 "errors" field, which may be either
```
natalie-o-perret: Nit: `7807` is made obsolete by`9457` tho, for this particular bit it is compliant with both, given that the PR aims at `9457`, I would use the latter.

v3/errors.go
```diff
@@ -94,11 +96,50 @@ var httpStatusCodeErrors = map[int]error{
 	http.StatusNetworkAuthenticationRequired: ErrNetworkAuthenticationRequired,
 }
 
+var jsonPathFieldRe = regexp.MustCompile(`\$\['([^']+)'\]`)
+
+// friendlyFieldName turns a JSONPath location like $['inference-engine-version']
+// into a CLI flag name like --inference-engine-version.
+func friendlyFieldName(location string) string {
```
natalie-o-perret: Unless I'm getting this wrong, this would add the `--` prefix in the SDK, this bakes CLI flag naming convention into egoscale. 

If this library is used in non-CLI contexts (other tools, services), that prefix is a wee confusing. 

This type presentation logic could arguably belong in the CLI layer, not the SDK, wdyt?

v3/errors.go
```diff
@@ -94,11 +96,50 @@ var httpStatusCodeErrors = map[int]error{
...
+		return location
+	}
+	return "--" + m[1]
+}
+
+// friendlyDetail rewrites technical validation messages into plain English.
+// "does not have a value in the enumeration ["a", "b"]" becomes
+// "invalid value; valid values are: a, b"
+func friendlyDetail(detail string) string {
+	const enumPrefix = "does not have a value in the enumeration "
```
natalie-o-perret: This parses enum values by splitting on `,`, and while it works for version strings, it would break if an enum value ever contained a comma (edge-case, probably yagni).

I think a more resilient alternative could be to parse the JSON array properly (`json.Unmarshal` into `[]string`) instead of string-splitting.

natalie-o-perret:
Lgtm, just a few nits that can be addressed imo, but not like huge blockers either.

natalie-o-perret (approved):

v3/errors.go
habfast: obsolete - i'm moving this to the CLI code. But good point, I'm picking up on it.

loicrouillermonay (approved):
LGTM

PR #765, opened 2026-04-24
fix(v3): derive zone endpoint from name without calling ListZones

natalie-o-perret:
## Problem

`GetZoneAPIEndpoint` and `GetZoneName` both called `ListZones` internally. 

That operation belongs to the compute API, so any IAM role without compute permissions would hit a `Forbidden` error the moment any v3-based command tried to route to a zone-specific endpoint.

The symptom reported in https://github.com/exoscale/cli/issues/824 is a DBaaS-only role failing on `exo dbaas update --valkey-ip-filter` with:

```
error: unable to create client: switch client zone v3: get zone api endpoint: list zones: ListZones: http response: Forbidden: Forbidden by role policy for compute
```

The same role works fine for MySQL and OpenSearch because older code paths bypass `SwitchClientZoneV3`. 

All newer service types (Valkey, Thanos, and anything added going forward) are affected.

## Fix

The zone API endpoint follows a stable, consistent URL pattern:

```
https://api-<zone-name>.exoscale.com/v2
```

We can derive the endpoint directly from the zone name without touching the API. 

The reverse direction (endpoint to zone name) is equally straightforward by trimming the known prefix and suffix.

This means `GetZoneAPIEndpoint` and `GetZoneName` no longer need any API call at all, so they work with any IAM role, regardless of compute permissions.

The generator template is updated alongside the generated code so future regenerations produce the same behaviour.

## Changes

- `v3/client.go`: replace both functions with string-based derivation, no API call
- `v3/generator/client/client.tmpl`: update the template to match
- `v3/client_test.go`: new test file covering all known zones, error cases, and a round-trip test

## Testing

```
go test ./v3/... -run TestGetZone -v
```

All 29 subtests pass. The full package test suite is also clean.

> [!NOTE] 
> GitHub Copilot with Claude Sonnet 4.6 was used for writing tests, and proofreading this description.

### DIFF
--- a/v3/client.go
+++ b/v3/client.go
@@ -32,6 +32,11 @@ const (
 	HrZag1 Endpoint = "https://api-hr-zag-1.exoscale.com/v2"
 )
 
+const (
+	zoneEndpointPrefix = "https://api-"
+	zoneEndpointSuffix = ".exoscale.com/v2"
+)
+
 // defaultHTTPClient is HTTP client with retry logic.
 // Default retry configuration can be found in go-retryablehttp repo.
 var defaultHTTPClient = func() *http.Client {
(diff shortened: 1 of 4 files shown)


kobajagi (changes requested):
This locks egoscale to production API.

natalie-o-perret:
Closing this in favor of fixing the root cause on the API side.

PR #766, opened 2026-04-27
v3: add AI API key operations and fix generator for RFC compliance

loicrouillermonay:
## Description

This PR ~adds AI API key management to egoscale v3 and~ fixes some v3 generator issues discovered while making it compatible with AI Services using [204 No Content](https://www.rfc-editor.org/rfc/rfc9110.html#name-204-no-content).

## Changes

- ~Add ListAIAPIKeys, CreateAIAPIKey, GetAIAPIKey, UpdateAIAPIKey, DeleteAIAPIKey, and RotateAIAPIKey~
- The generator crashed when rendering responses with no content (HTTP 204) because a nil check was missing. Then did a fix for generated functions signature because operations with no response body (like DELETE returning 204) now generate func (...) error instead of func (...) (*Type, error)
- Add ai acronym so generator produces AIAPIKey instead of AiApiKey

Bonus
- Every 4xx/5xx response leaks a TCP connection from the pool. The error handler read response bodies with `io.ReadAll` but never closed them. The fix is a one liner `defer resp.Body.Close()` inside the error-handling block.
- Generated error messages used "Json" instead of "JSON". [Go conventions says: consistent case for acronyms.](https://go.dev/wiki/CodeReviewComments#initialisms)

The connection leak and the JSON naming fix are outside the scope of my task which is to add support for the AI API key operations but I included them because they are minor one-line changes. I can revert those commits if you prefer

## Checklist

- [x] Changelog updated (under *Unreleased* block)

### DIFF
--- a/v3/errors.go
+++ b/v3/errors.go
@@ -118,14 +118,13 @@ type APIErrorEntry struct {
 	Location string
 }
 
-func (e *APIError) Error() string {
-	return fmt.Sprintf("%s: %s", e.sentinel.Error(), e.Message)
-}
-
+func (e *APIError) Error() string { return e.sentinel.Error() }
 func (e *APIError) Unwrap() error { return e.sentinel }
 
 func handleHTTPErrorResp(resp *http.Response) error {
 	if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
+		defer resp.Body.Close() // prevent connection leak on error responses
+
 		var res struct {
 			Message string          `json:"message"`
 			Error   string          `json:"error"`
--- a/v3/operations.go
+++ b/v3/operations.go
@@ -69,7 +69,7 @@ func (c Client) ListAIAPIKeys(ctx context.Context) (*ListAIAPIKeysResponse, erro
 
 	bodyresp := new(ListAIAPIKeysResponse)
 	if err := prepareJSONResponse(response, bodyresp); err != nil {
-		return nil, fmt.Errorf("ListAIAPIKeys: prepare Json response: %w", err)
+		return nil, fmt.Errorf("ListAIAPIKeys: prepare JSON response: %w", err)
 	}
 
 	return bodyresp, nil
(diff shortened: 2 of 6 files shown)


pierre-emmanuelJ (approved):

v3/generator/source.yaml
kobajagi, line 1: Can we remove the spec bump to be handled by our regular cron action?
Just leave generator code updates and you can start spec action manually (after merge) if you don't want to wait:
https://github.com/exoscale/egoscale/actions/workflows/generate_v3.yaml

kobajagi (changes requested):

loicrouillermonay:
Thanks for the good feedback @kobajagi, the PR now only contains the fixes and the AI API key operations comes from the github workflow script

aureliar8:
FYI for VPC we are also interested in HTTP 204 support

v3/errors.go
```diff
@@ -123,6 +123,8 @@ func (e *APIError) Unwrap() error { return e.sentinel }
 
 func handleHTTPErrorResp(resp *http.Response) error {
 	if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
+		defer func() { _ = resp.Body.Close() }() // prevent connection leak on error responses
```
aureliar8:
```suggestion
		defer resp.Body.Close() // prevent connection leak on error responses
```

I know that people often do that to silence linters but I'd argue that linters should be correctly configured to not warn of this very fine pattern

loicrouillermonay:
@aureliar8 that's good to know, in the meantime in AI-services I reverted to not using HTTP 204 for now to move forward more quickly.

I'll let @exoscale/giza and @kobajagi confirm they're ok with most changes from this PR. On my side most of the things I thought I needed or wanted were reverted and it's really just about suggesting all the small improvements that were noticed along the way.

aureliar8:
> @aureliar8 that's good to know, in the meantime in AI-services I reverted to not using HTTP 204 for now to move forward more quickly.

Yeah it's the same for us, to move quickly we are returning 200 OK with an empty JSON body, but the generated code is uglier than if 204 were really supported

v3/operations.go
quentinalbertone, line 1: **nitpick:** please do not edit this file

quentinalbertone (changes requested):

v3/operations.go
loicrouillermonay: But if I don't do it the CI won't pass since I'm suggesting using "Json" instead of "JSON" to follow [Go conventions](https://go.dev/wiki/CodeReviewComments#initialisms) so I felt like I need to regenerate this manually in the context of this PR. No?

aureliar8:
[ch189404]

v3/operations.go
kobajagi: It is fine to run the generator and commit changes in `operations.go`. Just don't bump openapi spec.

kobajagi: All good here from my POV

kobajagi (approved):

PR #773, opened 2026-05-21
v3/generator: fall back to string for unknown schema formats

natalie-o-perret:
Formats that are not valid Go identifiers (e.g. `uri-reference`) were returned verbatim from `RenderSimpleType`, producing uncompilable output.

Fall back to `string` for anything not explicitly handled. Also adds an explicit case for the `int64` format so it keeps mapping to `int64` instead of falling through to the new default.

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -108,17 +108,25 @@ func RenderSimpleType(s *base.Schema) string {
 	}
 
 	if s.Format != "" {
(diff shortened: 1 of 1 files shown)


natalie-o-perret:
Superseded - master already has dedicated fixes for uri-reference (#753) and byte (#759). Our branch was also causing CI failures because it was based on an old commit (pre-v3.1.34/35/36 tags), which made go get downgrade egoscale for downstream builds.

PR #775, opened 2026-05-22
v3/generator: handle OAS 3.1 nullable types

natalie-o-perret:
OAS 3.1 expresses nullability as a type array rather than a boolean flag,  e.g. `["integer", "null"]` instead of `nullable: true`. 

3 sites in the generator assumed `Type[0]` is always the real type, which breaks when `null` appears first.

## Changes

- **`RenderSimpleType`**, iterate `Type`, skip `"null"`, use the first real entry instead of `Type[0]`
- **`IsSimpleSchema`**, same, skip `"null"` before checking for `object`/`map`/`array`
- **`renderObject`**, detect `"null"` in the type array and set`nullable = true`, so a field typed `["integer", "null"]` generates `*int` instead of `int`

The current `source.yaml` still uses OAS 3.0 `nullable: true` throughout, so generated output is unchanged. This is a forward-looking fix for when the spec migrates to 3.1.

## Tests

All existing tests pass, no regressions.

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -108,17 +108,25 @@ func RenderSimpleType(s *base.Schema) string {
 	}
 
 	if s.Format != "" {
-		if s.Format == "date-time" {
+		switch s.Format {
+		case "date-time":
 			return "time.Time"
-		}
-		if s.Format == "uuid" {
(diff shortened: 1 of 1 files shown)


natalie-o-perret:
Superseded by new branch based on current master.

PR #776, opened 2026-05-22
v3/generator: handle OAS 3.1 nullable type arrays

natalie-o-perret:
# Description

OpenAPI 3.1 expresses nullable values as a type array, such as `["integer", "null"]`. The v3 generator did not mark these object properties nullable, so it generated `int` instead of `*int`.

This change:

- handles exactly one concrete property type plus `null`
- rejects heterogeneous and null-only type arrays instead of silently choosing a type
- regenerates the affected quota request and response fields as `*int`
- adds focused regression tests for both type orderings and unsupported arrays

## Checklist
(For exoscale contributors)

---

> [!NOTE]
> AI assistance: test scaffolding.

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -119,9 +119,18 @@ func RenderSimpleType(s *base.Schema) string {
 			return "string"
 		case "byte":
 			return "[]byte"
-		default:
-			return s.Format
+		case "int32":
+			return "int32"
+		case "int64":
+			return "int64"
+		case "float":
+			return "float32"
+		case "double":
+			return "float64"
 		}
+		// Any other format is still a string at the Go level;
+		// the format is just a validation hint.
+		return "string"
 	}
 
 	if len(s.Type) == 0 {
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -101,9 +101,9 @@ func RenderSchema(schemaName string, s *base.SchemaProxy) ([]byte, error) {
 // Add more simple type here.
 func RenderSimpleType(s *base.Schema) string {
 	if s.Extensions != nil {
-		typ, ok := s.Extensions.Get("x-go-type")
+		goType, ok := s.Extensions.Get("x-go-type")
 		if ok {
-			return typ.Value
+			return goType.Value
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


v3/generator/schemas/schemas.go
```diff
@@ -132,17 +141,27 @@ func RenderSimpleType(s *base.Schema) string {
...
-	if s.Type[0] == "boolean" {
+
+	// OAS 3.1: type can be ["integer", "null"], skip null to get the real type.
+	kind := ""
+	for _, entry := range s.Type {
+		if entry != "null" {
+			kind = entry
+			break
+		}
+	}
```
imiric: Hhmm this picks the first non-null type it finds, but AFAIU this is a union field that could specify more than 1 type, so what should be the correct return value in that case? :thinking:

Possibly `any`, no? And even for `["null"]`, should it still be `any`?

I think there's another issue in that [according to the JSON Schema spec the value of `type` could also be a string](https://json-schema.org/draft/2020-12/json-schema-validation#name-type), but that doesn't seem supported by the `pb33f/libopenapi` dep we're using, so I suppose we can't handle that scenario anyway.

imiric (commented):
This is definitely an improvement, but I'm not sure about the `type` handling. The previous code had the same issue, so maybe it doesn't need to be fixed here(?).

Keep in mind that my familiarity with egoscale and OpenAPI is low, so feel free to push back 🙂

pierre-emmanuelJ (approved):

kobajagi:
Sorry, this PR went below radar. Please rebase it and i'll review it ASAP.

natalie-o-perret:
> Sorry, this PR went below radar. Please rebase it and i'll review it ASAP.

My bad, it equally went under my own radar :woman_facepalming: :grimacing:, will take care of that today.

v3/generator/schemas/schemas.go
natalie-o-perret: Good catch. I've removed the 1st-non-null approach.

The newly rebased version only handles one concrete type plus `null`. 

Real unions and `["null"]` now fail generation instead of becoming `any` or silently picking a type.

Also, `libopenapi` supports scalar `type` values and "normalizes" both forms to `[]string`.

natalie-o-perret:
> Sorry, this PR went below radar. Please rebase it and i'll review it ASAP.

@kobajagi, it has been rebased and adjusted accordingly :bowing_woman:

kobajagi (approved):

imiric (approved):

PR #787, opened 2026-06-22
fix(generator): nvidia-mig-specific enum collision

PhilippeChepy:
# Description

Since we don't have any other use case and we need to keep backward compatibility on pre-existing symbols, it's scoped to Nvidia MIG symbols only.

## Checklist
(For exoscale contributors)

### DIFF
--- a/v3/generator/schemas/schemas.go
+++ b/v3/generator/schemas/schemas.go
@@ -233,13 +233,27 @@ func renderSimpleTypeEnum(typeName string, s *base.Schema) string {
 		if typ == "string" {
 			value = fmt.Sprintf("%q", e.Value)
 		}
-		name := typeName + helpers.ToCamel(e.Value)
+
+		suffix := e.Value
+
+		// Scope "+"/"-" enum sign encoding to NVIDIA MIG profiles.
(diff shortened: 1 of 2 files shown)


v3/generator/schemas/schemas.go
```diff
@@ -233,7 +233,14 @@ func renderSimpleTypeEnum(typeName string, s *base.Schema) string {
 		if typ == "string" {
 			value = fmt.Sprintf("%q", e.Value)
 		}
-		name := typeName + helpers.ToCamel(e.Value)
+
+		suffix := e.Value
+
+		// Scope "+"/"-" enum sign encoding to NVIDIA MIG profiles.
+		if strings.HasPrefix(typeName, "NvidiaMig") {
```
kobajagi, line 240: Solves the problem but is very much a ducktape and can bite us back later.

kobajagi (changes requested):
Let us explore the alternatives first and keep this as a last resort.

PhilippeChepy:
likely superseded by https://github.com/exoscale/egoscale/pull/789

quentinalbertone:
done by PR https://github.com/exoscale/egoscale/pull/789

PR #806, opened 2026-08-27
v2: preserve custom API endpoints

natalie-o-perret:
# Description

Preserve custom v2 API endpoints instead of rewriting hostname endpoints to zonal Exoscale hosts. Default and environment endpoints keep their zonal routing.

Stacked on #807 and related to [terraform-provider-exoscale#573](https://github.com/exoscale/terraform-provider-exoscale/issues/573).

## Checklist
(For exoscale contributors)

### DIFF
--- a/v2/client.go
+++ b/v2/client.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"net"
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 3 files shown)


natalie-o-perret:
[SC-193400]

v2/client.go
```diff
@@ -254,22 +253,23 @@ func setUserAgent(ctx context.Context, req *http.Request) error {
...
 	}
-	if net.ParseIP(h) == nil {
-		v, ok := ctx.Value(api.ReqEndpoint{}).(api.ReqEndpoint)
-		if ok {
-			req.Host = v.Host()
-			req.URL.Host = v.Host()
-		}
+
+	origin := req.URL.Scheme + "://" + req.URL.Host + "/"
+	if origin != api.EndpointURL && origin != "https://"+v.Env()+".exoscale.com/" {
```
kobajagi, line 266: Why is second clause needed?

natalie-o-perret: The 1st clause only covers the default `api` endpoint. The 2nd preserves the existing zonal routing for other API environments: `https://<env>.exoscale.com/` must still become `<env>-<zone>.exoscale.com`. Without it, those endpoints would be incorrectly treated as custom. The `environment endpoint` test covers this case.

pierre-emmanuelJ (approved):

kobajagi (approved):

PR #807, opened 2026-08-27
lint: ignore deprecated v1 assignments

natalie-o-perret:
# Description

Ignore `SA4006` in the deprecated v1 client test while keeping Staticcheck enabled for v2. This leaves the dead v1 code unchanged and unblocks CI for #806.

## Checklist
(For exoscale contributors)

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -550,16 +550,6 @@ func TestClientList(t *testing.T) {
 	}}`
 
 	for _, tt := range lsTests() {
-		end := len(tt.name)
-		switch {
-		case strings.HasSuffix(tt.name, "ses"):
-			end -= 2 // nolint: ineffassign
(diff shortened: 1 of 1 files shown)


kobajagi (changes requested):
This file is a dead code as is all code in the root. This was egoscale v1 which targets API v1 which was decommissioned years ago. I would leave it as is and setup linter to ignore code outside of v3. Eventually we should do the full cleanup of v1 code.

kobajagi (approved):

natalie-o-perret:
Thanks, agreed. I've restored the deprecated v1 test code and scoped the Staticcheck ignore to SA4006 in that file only. v2 remains covered by the existing `./...` lint target.

